PR #24, opened 2024-09-19
Add missing property redirected to VisitResponse

olivaresf:
Turbo-iOS PR: https://github.com/hotwired/turbo-ios/pull/225

> When a redirection with a responseHTML would happen (e.g. a form submission), Turbo Native receives something like the following:
> 
> ```
> ← visitProposed [
>     "location": http://localhost:45678/submit_redirected, 
>     "timestamp": 1725557992593, 
>     "options": {
>         action = replace;
>         response =     {
>             redirected = 1;
>             responseHTML = "{HTML with flash notice}";
>             statusCode = 200;
>         };
>         shouldCacheSnapshot = 0;
>     }
> ]
> ```
> 
> This usually gets turned into a JavaScript visit and then handed over to the bridge, which begins the visit by calling `visitLocationWithOptionsAndRestorationIdentifier`.
> 
> However, when transforming this proposal JS -> Swift (for processing) and back to JS (for bridge consumption), `visitProposed["options"]["response"]["redirected"]` was being lost. This caused `responseHTML` to be ignored and a replace visit to be proposed instead.
> 
> By adding `redirected` to `VisitResponse` the behavior is now correct.


jayohms:
Closing this in favor of: https://github.com/hotwired/hotwire-native-ios/pull/60

PR #36, opened 2024-10-10
Fix authentication flow in demo app

joemasilotti:
When migrating from turbo-ios, `visitableDidFailRequest(_:error:retryHandler:)` was left behind. This PR adds that back in which fixes the authentication flow when hitting a 401.


zoejessica (approved):
Looks great! 👍

PR #37, opened 2024-10-11
fix: VisitableView NavigationBar glitch

CopperGiraffe:
Hello,

I believe this is fixing the NavigationBar glitches described in https://github.com/hotwired/turbo-ios/issues/196.
It’s still not perfect when using large titles but otherwise it seems to do the trick.

This PR is just changing the way we insert Subviews in the VisitableView.swift file to make sure they are always inserted the way we want.


arjun810:
Seeing this bug on the demo app made me a bit nervous about the state of hotwire native, so it'd be nice to fix this. I can't comment on the fix in this PR, but fixing the demo app at a minimum would be great.

joemasilotti:
This will be fixed with #87.

PR #42, opened 2024-10-21
Fix `animated` flag from path configuration

joemasilotti:
This PR addresses a mistake noticed by @mrfidgety in the [turbo-ios repo](https://github.com/hotwired/turbo-ios/pull/158#pullrequestreview-2329794028). We were looking for the `animated` flag in the `parameters` but it should appear in `properties` because it comes from the path configuration.

This PR fixes that typo.


svara (approved):
Good catch!

PR #43, opened 2024-11-08
Expose `Session.reload()` on Navigator

joemasilotti:
This PR exposes `reload()` on the main `Session`. When working with tabs, it's helpful to be able to manually reload screens after the user signs in or out.

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -76,9 +76,10 @@ public class Navigator {
         hierarchyController.clearAll(animated: animated)
     }
 
-    /// Reloads the main `Session`.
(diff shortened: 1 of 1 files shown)


Source/Turbo/Navigator/Navigator.swift
```diff
@@ -76,6 +76,11 @@ public class Navigator {
         hierarchyController.clearAll(animated: animated)
     }
 
+    /// Reloads the main `Session`.
+    public func reload() {
```
svara, line 80: I'm wondering if we should reload the modal session too.

joemasilotti: Yeah good call - will do.

PR #55, opened 2024-12-06
Global configuration consolidation

svara:
This PR consolidates the global configuration.

**Changes**
- `PathConfiguration` is now global.
- All global config in the `Demo` app has been moved to `AppDelegate`.
- `Navigator` automatically sets the session's path config from the global path config.

### DIFF
--- a/Demo/SceneController.swift
+++ b/Demo/SceneController.swift
@@ -7,17 +7,7 @@ final class SceneController: UIResponder {
     var window: UIWindow?
 
     private let rootURL = Demo.current
-    private lazy var navigator = Navigator(pathConfiguration: pathConfiguration, delegate: self)
-
-    // MARK: - Setup
-
-    private func configureRootViewController() {
-        guard let window = window else {
-            fatalError()
-        }
-
-        window.rootViewController = navigator.rootViewController
-    }
+    private lazy var navigator = Navigator(delegate: self)
 
     // MARK: - Authentication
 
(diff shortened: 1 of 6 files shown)


Demo/SceneController.swift
```diff
@@ -7,50 +7,24 @@ final class SceneController: UIResponder {
...
-    }
-
-    private func configureRootViewController() {
-        guard let window = window else {
-            fatalError()
-        }
-
-        window.rootViewController = navigator.rootViewController
-    }
+    private lazy var navigator = Navigator(delegate: self)
```
jayohms, line 10: This no longer needs to be `lazy`, correct?

jayohms (approved):
Very nice @svara 👍. This can be part of an upcoming `v1.1` since it changes a couple public APIs. We can merge this once the corresponding site documentation PR is ready, too.

@joemasilotti let us know if you spot any concerns.

Demo/SceneController.swift
joemasilotti: It still needs to be `lazy` because we are referencing `self` before `self` is instantiated.

Source/HotwireConfig.swift
```diff
@@ -78,11 +78,3 @@ public struct HotwireConfig {
         return configuration
     }
 }
-
-public extension HotwireConfig {
-    class PathConfiguration {
-        /// Enable to include the query string (in addition to the path) when applying rules.
-        /// Disable to only consider the path when applying rules.
-        public var matchQueryStrings = true
```
joemasilotti, line 86: Why did this move out of the configuration? How can I access it if the new variable isn't `static`?

I'd vote to keep this inside of the global configuration if possible.

joemasilotti: Disregard my comment above. The diff doesn't show that `Hotwire.config.pathConfiguration` still exists. The location of `matchQueryStrings` just moved.

joemasilotti (approved):

PR #56, opened 2024-12-12
Breaking API change: applications can set a User-Agent prefix

joemasilotti:
Brings [Android PR #77](https://github.com/hotwired/hotwire-native-android/pull/77) to iOS, allowing developers to set a custom user agent *prefix* without breaking bridge components or removing "Hotwire Native".

Developers can set a custom `applicationUserAgentPrefix` for their application via:
```swift
Hotwire.config.applicationUserAgentPrefix = "My Application Prefix;"
```

Whenever a `WKWebView` instance is created, the library will now automatically include:
- Your (optional) custom `applicationUserAgentPrefix`
- "Hotwire Native iOS; Turbo Native iOS;"
- "bridge-components: [your bridge components];"

`WKWebView`'s default user agent will also be added to the beginning of the user agent.

### DIFF
--- a/Demo/SceneController.swift
+++ b/Demo/SceneController.swift
@@ -12,6 +12,10 @@ final class SceneController: UIResponder {
     // MARK: - Setup
 
     private func configureBridge() {
+        // TODO: Move to AppDelegate when PR #55 is merged.
+        // https://github.com/hotwired/hotwire-native-ios/pull/55/files
+        Hotwire.config.applicationUserAgentPrefix = "Hotwire Demo;"
+
         Hotwire.registerBridgeComponents([
             FormComponent.self,
             MenuComponent.self,
--- a/Source/Bridge/UserAgent.swift
+++ b/Source/Bridge/UserAgent.swift
@@ -1,8 +1,15 @@
 import Foundation
 
-public enum UserAgent {
-    public static func userAgentSubstring(for componentTypes: [BridgeComponent.Type]) -> String {
+enum UserAgent {
+    static func build(componentTypes: [BridgeComponent.Type], applicationPrefix: String?) -> String {
         let components = componentTypes.map { $0.name }.joined(separator: " ")
-        return "bridge-components: [\(components)]"
+        let componentsSubstring = "bridge-components: [\(components)]"
+
+        return [
+            applicationPrefix,
+            "Hotwire Native iOS;",
+            "Turbo Native iOS;",
+            componentsSubstring
+        ].compactMap { $0 }.joined(separator: " ")
     }
 }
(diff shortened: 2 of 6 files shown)


Source/Bridge/UserAgent.swift
```diff
@@ -1,8 +1,15 @@
 import Foundation
 
-public enum UserAgent {
-    public static func userAgentSubstring(for componentTypes: [BridgeComponent.Type]) -> String {
+enum UserAgent {
+    static func build(componentTypes: [BridgeComponent.Type], applicationPrefix: String?) -> String {
```
joemasilotti, line 4: I'm a bit worried about this change since it was previously public. But I don't think we need/want it to be anymore. Since we are bumping to v1.1 I think that should be OK - I like hiding this from developers.

Tests/Bridge/UserAgentTests.swift
```diff
@@ -1,15 +1,29 @@
 import Foundation
-import HotwireNative
+@testable import HotwireNative
```
joemasilotti, line 2: `@testable` is needed to access the (now) internal `UserAgent` function.

Demo/SceneController.swift
```diff
@@ -12,6 +12,10 @@ final class SceneController: UIResponder {
     // MARK: - Setup
 
     private func configureBridge() {
+        // TODO: Move to AppDelegate when PR #55 is merged.
+        // https://github.com/hotwired/hotwire-native-ios/pull/55/files
+        Hotwire.config.applicationUserAgentPrefix = "Hotwire Demo;"
```
joemasilotti, line 17: Commenting so we don't forget about this to-do.

Source/Bridge/UserAgent.swift
jayohms: I'm not worried about this, since we're already breaking the API and there's no application need to build the user agent manually. Can we flip the param order since the `applicationPrefix` will be first in the user agent?

Source/Bridge/UserAgent.swift
```diff
@@ -1,8 +1,15 @@
...
         let components = componentTypes.map { $0.name }.joined(separator: " ")
-        return "bridge-components: [\(components)]"
+        let componentsSubstring = "bridge-components: [\(components)]"
+
+        return [
+            applicationPrefix,
+            "Hotwire Native iOS;",
+            "Turbo Native iOS;",
+            componentsSubstring
+        ].compactMap { $0 }.joined(separator: " ")
```
jayohms, line 13: Nice implementation, I may do that on Android too 👍

jayohms (approved):
Thanks!

Source/Bridge/UserAgent.swift
joemasilotti: Done!

svara (approved):
Nicely done @joemasilotti 👏🏻

Issue #57, opened 2024-12-14
`Hotwire.config.makeCustomWebView` conflicts with `Hotwire.registerBridgeComponents()`


joemasilotti:
Using both `Hotwire.config.makeCustomWebView` and `Hotwire.registerBridgeComponents()` causes unexpected behavior. If the web view is configured first then it is not used. If bridge components are registered first then the bridge is not initialized in the web view.

It's not uncommon for developers to need to do both. Disabling link previews, for example, requires access to the underlying web view.

```swift
Hotwire.config.makeCustomWebView = { config in
    let webView = WKWebView(frame: .zero, configuration: config)
    webView.allowsLinkPreview = false
    return webView
}
```

Currently, a developer must choose between bridge components *or* a custom web view.

This is happening because `Hotwire.registerBridgeComponents()` overwrites `Hotwire.config.makeCustomWebView` to initialize the bridge.

PR #58, opened 2024-12-15
Document makeCustomWebView w/ bridge components

joemasilotti:
As discussed in #57:

> Using both `Hotwire.config.makeCustomWebView` and `Hotwire.registerBridgeComponents()` causes unexpected behavior. If the web view is configured first then it is not used. If bridge components are registered first then the bridge is not initialized in the web view.

This PR documents that a developer must manually call `Bridge.initialize(webView)` if using `Hotwire.config.makeCustomWebView` *and* call it after calling `Hotwire.registerBridgeComponents()`. It's not ideal, but I'm not sure of a better fix just yet.

But we do have some nice option-click documentation now!

image

### DIFF
--- a/Source/HotwireConfig.swift
+++ b/Source/HotwireConfig.swift
@@ -42,7 +42,19 @@ public struct HotwireConfig {
     }
 
     /// Optionally customize the web views used by each Turbo Session.
+    ///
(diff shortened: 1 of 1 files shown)


jayohms:
Closing since we have an incoming fix.

PR #61, opened 2024-12-17
Additional modal presentation styles

svara:
This PR adds two modal styles: page sheet and form sheet. These styles differ in presentation only on iPads. On iPhones, they fallback to the same behavior:
• In portrait mode, the system displays the view controller as a sheet, with part of the background content visible near the top of the screen.
• In landscape mode, the behavior is identical to full screen presentation.

| Page sheet | Form sheet |
|--------|--------|
| page sheet | form sheet | 

Additionally, this PR also adds the option to disable the pull-down gesture for the page and form sheet presentations. This is particularly important on iPads, where all it takes is an accidental tap outside of the sheet to dismiss it.
You can specify the behaviour by adding `modal_sheet_dismiss_gesture_enabled` to the path config. If omitted, the default value `true` is used, which matches the default OS behaviour.
```
{
   "rules":[
      {
         "patterns":[
            "/recipes/*"
         ],
         "properties":{
            "context": "modal",
            "modal_sheet_dismiss_gesture_enabled": false,
            "view_controller":"recipes"
         }
      }
   ]
}
```


Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
```diff
@@ -26,6 +26,14 @@ public extension VisitProposal {
         properties["pull_to_refresh_enabled"] as? Bool ?? true
     }
 
+    var modalSheetDismissGestureEnabled: Bool {
+        if let dismissEnabled = properties["modal_sheet_dismiss_gesture_enabled"] as? Bool {
```
jayohms, line 30: Since this gesture applies to all modal styles (not just the `pageSheet` and `formSheet` styles), maybe it's be more clear and less verbose to drop the `_sheet_` part of the name and go with:
```
modal_dismiss_gesture_enabled
```

What do you think?

jayohms (approved):
Awesome, thanks!

Source/Turbo/Navigator/Helpers/Navigation.swift
```diff
@@ -18,5 +18,7 @@ public enum Navigation {
         case medium
         case large
         case full
+        case pageSheet
+        case formSheet
```
joemasilotti, line 22: I hate to come in late after a merge, but I think these should be:

```swift
case pageSheet = "page_sheet"
case formSheet = "form_sheet"
```

to match `Navigation.Presentation.clearAll` and `Navigation.Presentation.replaceRoot`.

PR #64, opened 2024-12-21
add public method to allow clearing the turbo session cache

yanshiyason:
I'm encountering an issue with my `UITabBarController`-based app related to navigation state and Turbo session cache behavior.

#### Scenario

- **Tab A** Navigation Stack:
  1. `/posts` (index page)
  2. `/posts/1` (show page)

- **Tab B** Navigation Stack:
  1. `/posts` (index page)
  2. `/posts/1` (show page)
  3. `/posts/1/edit` (edit page, where a POST request is made)
  4. Back to `/posts/1` (show page)  
     - The Turbo session cache is **busted**, and the back button now shows updated data on the index page.

#### Issue

After interacting with Tab B and returning to Tab A:

1. Tab A’s `/posts/1` (show page) is updated as expected because the navigator executed a `Turbo.visit(page.url, {action: 'replace'})`.
2. However, when pressing the back button in Tab A:
   - The `/posts` (index page) is **stale** because the Turbo session cache for Tab A’s navigator was **not busted**.

Having a way to clear the cache manually would fix my issue.

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -87,6 +87,11 @@ public class Navigator {
         modalSession.reload()
     }
 
+    /// Invalidates the session cache.
+    public func invalidateSessionCache() {
+        session.markSnapshotCacheAsStale()
+    }
+
     /// Navigate to an external URL.
     ///
     /// - Parameters:


joemasilotti:
We added `reload()` to handle this problem in a slightly different way. 

https://github.com/hotwired/hotwire-native-ios/blob/aa5da14014143be76e7856f2bf07cf887fb7bfe0/Source/Turbo/Navigator/Navigator.swift#L85

Reloading ensures that the current page stays up to date, too. Not just when you tap the back button. 

I’m curious to hear if this solves your problem or you need to only clear the cache.

yanshiyason:
Hi Joe, thank you for getting back to me so quickly.

So I tried both approaches, and they don't behave the same. Here is what I found:

```
Navigation stack:
1. /home
2. /home/top-posts
3. /home/top-posts/1

navigator.reload()

Navigation stack:
1. /home
2. /home/top-posts
3. /home           <-- This changed from current URL to initial URL
```

```
Navigation stack:
1. /home
2. /home/top-posts
3. /home/top-posts/1

navigator.invalidateSessionCache()
navigator.route(VisitProposal(url: url, options: VisitOptions(action: .replace)))

1. /home
2. /home/top-posts
3. /home/top-posts/1  <-- This is still the current URL
```

So apparently, `session.reload()` loads the URL the session was initially loaded with, not the current URL in the webView.

joemasilotti:
Hmm, that doesn't seem correct. I can't reproduce in the demo app. I added a button to call `navigator.reload()` in the demo app and it kept the current stack.

Are you sure there isn't anything going on on your server to redirect a page?

If you can reproduce in the demo app please let me know.

yanshiyason:
You are correct! I just tested again and it seems to work correctly using navigator.reload().
Sorry for the false alert.

Issue #12, opened 2024-04-06
Double page requests when interacting with modals


joemasilotti:
In merging outstanding PRs from turbo-ios I learned that [PR #155](https://github.com/hotwired/turbo-ios/pull/155) doesn't play nicely with Turbo Navigator. It can cause screens to not render at all due to the cached visit options being out of sync. I decided to skip the merge (for now) because it was causing more issues than it fixed.

A new, and I think related, issue has popped up instead. Using the Demo app, double requests are not being made to anything related to modals.

Launching the demo app and:
1. Tapping "Load a webpage modally" requests `/new` twice
2. Submitting the form POSTs once correctly but requests `/success` twice

olivaresf:
This should be fixed in https://github.com/hotwired/hotwire-native-ios/pull/24.

olivaresf:
@joemasilotti could you help me review #24? I believe that should fix the issue.

joemasilotti:
I'm still seeing the same behavior of modals and form submissions hitting twice in the demo app (and same for my Rails server):

```
Mon Oct 21 2024 11:09:00 GMT-0700 (Pacific Daylight Time) -- GET /new
Mon Oct 21 2024 11:09:00 GMT-0700 (Pacific Daylight Time) -- GET /new
Mon Oct 21 2024 11:09:01 GMT-0700 (Pacific Daylight Time) -- POST /new
Mon Oct 21 2024 11:09:01 GMT-0700 (Pacific Daylight Time) -- GET /success
Mon Oct 21 2024 11:09:01 GMT-0700 (Pacific Daylight Time) -- GET /success
```

jayohms:
Note, this issue is due to two completely separate issues:

1. Double-requests to `/new` is due to Turbo link prefetching being initiated when tapping on the `/new` link, but prefetching results won't work across multiple `default`/`modal` `WKWebView` sessions. To address this, we plan to disable prefetching in the native libraries, since link taps was not an expected use of prefetching anyway (as opposed to mouse hover events).
2. Double-requests to `/success` is due to the issue outlined in this PR: #60

jayohms:
Both issues were resolved with:

- https://github.com/hotwired/hotwire-native-ios/pull/60
- https://github.com/hotwired/hotwire-native-ios/pull/62

PR #70, opened 2025-01-09
Detect cross-origin redirects during visits

svara:
This PR is the counterpart of the Android https://github.com/hotwired/hotwire-native-android/pull/82. 
Due to CORS restrictions, the javascript Fetch API cannot be reliably used across applications to detect cross-origin redirects during Turbo visits.

This new approach hands off failed visits with a non-HTTP status code (which suggests a cross-origin redirect may have been attempted) to the native code to detect cross-origin redirects. If a visit redirect is detected, the final redirect location is treated as an external URL and opened appropriately.

### DIFF
--- a/Source/Turbo/Networking/RedirectHandler.swift
+++ b/Source/Turbo/Networking/RedirectHandler.swift
@@ -15,7 +15,7 @@ enum RedirectHandlerError: Error {
 struct RedirectHandler {
     enum Result {
         case noRedirect
-        case redirect(URL) // Same-domain redirect.
+        case sameOriginRedirect(URL)
         case crossOriginRedirect(URL)
     }
 
@@ -40,7 +40,7 @@ struct RedirectHandler {
                 return .crossOriginRedirect(responseUrl)
             }
 
-            return .redirect(responseUrl)
+            return .sameOriginRedirect(responseUrl)
         } catch let error as RedirectHandlerError {
             throw error
         } catch {
--- a/Source/Turbo/Networking/RedirectHandler.swift
+++ b/Source/Turbo/Networking/RedirectHandler.swift
@@ -1,0 +1,65 @@
+import Foundation
+
+enum RedirectHandlerError: Error {
+    case requestFailed(Error)
+    case responseValidationFailed(reason: ResponseValidationFailureReason)
+
+    /// The underlying reason the `.responseValidationFailed` error occurred.
+    public enum ResponseValidationFailureReason: Sendable {
+        case missingURL
+        case invalidResponse
+        case unacceptableStatusCode(code: Int)
+    }
+}
+
+struct RedirectHandler {
+    enum Result {
+        case noRedirect
+        case redirect(URL)
+        case crossOriginRedirect(URL)
+    }
+
+    func resolve(location: URL) async throws -> Result {
+        do {
+            let request = URLRequest(url: location)
+            let (data, response) = try await URLSession.shared.data(for: request)
+            guard let httpResponse = response as? HTTPURLResponse else {
+                throw RedirectHandlerError.responseValidationFailed(reason: .invalidResponse)
+            }
+
+            guard httpResponse.isSuccessful else {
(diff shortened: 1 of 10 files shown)


Source/Turbo/Networking/RedirectHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        return httpResponse
+    }
+}
+
+extension HTTPURLResponse {
+    public var isSuccessful: Bool {
+        switch statusCode {
+        case 200 ... 299:
```
jayohms, line 67: Note: HTTP `3xx` status codes are typically considered logically successful responses, but since we're automatically following redirects in the request, we should never get into a situation where we end on a `3xx` status code. Fine to check for `2xx` 👍

Source/Turbo/Visit/ColdBootVisit.swift
```diff
@@ -67,9 +67,24 @@ extension ColdBootVisit: WKNavigationDelegate {
...
+            decisionHandler(.cancel)
+            return
+        }
+
+        let isRedirect = location != url
+        let redirectIsCrossOrigin = isRedirect && location.host != url.host
+
+        if redirectIsCrossOrigin {
+            decisionHandler(.cancel)
+            UIApplication.shared.open(url)
```
jayohms, line 83: Can we log the cold boot cross-origin redirect here, so it's easy to understand in logs?

Source/Turbo/Networking/RedirectHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+        case missingURL
+        case invalidResponse
+        case unacceptableStatusCode(code: Int)
+    }
+}
+
+struct RedirectHandler {
+    enum Result {
+        case noRedirect
+        case redirect(URL) // Same-domain redirect.
```
jayohms, line 18: Maybe call this `sameOriginRedirect()` for clarity?

Source/Turbo/Session/SessionDelegate.swift
```diff
@@ -2,6 +2,7 @@ import UIKit
 
 public protocol SessionDelegate: AnyObject {
     func session(_ session: Session, didProposeVisit proposal: VisitProposal)
+    func session(_ session:Session,  didProposeVisitToCrossOriginRedirect location: URL)
```
jayohms, line 5: Weird spacing issue.

jayohms (approved):
Looks great, nice work on this @svara.

Source/Turbo/Networking/RedirectHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+        guard httpResponse.isSuccessful else {
+            throw RedirectHandlerError.responseValidationFailed(reason: .unacceptableStatusCode(code: httpResponse.statusCode))
+        }
+
+        return httpResponse
+    }
+}
+
+extension HTTPURLResponse {
+    public var isSuccessful: Bool {
```
joemasilotti, line 65:
```suggestion
    public var isSuccessful: Bool { (200...299).contains(statusCode) }
```

joemasilotti:
Using [Demo PR #12](https://github.com/hotwired/hotwire-native-demo/pull/12), I'm seeing the external redirect open in `SFSafariViewController`. What do I need to configure to test this fully? I'm not seeing how I can implement `handle(externalURL:)` for this to work.

Source/Turbo/Networking/RedirectHandler.swift
svara: Correct.

Source/Turbo/Visit/ColdBootVisit.swift
svara: Added.

ibrahima:
Hi! This is helpful, I've noticed this issue before as well, so thanks for resolving it!

I was wondering, is there a way to handle this differently from within my application code? E.g., if I wanted to open this in a webview within my app, would there be a way to set that up? (At this point I don't know for sure if this would help me, but it's something I'm considering to handle a specific situation with my app.)

Issue #71, opened 2025-01-15
add Server-Driven Routes to path configurations?


subelsky:
I discovered these useful `turbo-rails` route helpers in the [Hotwire docs](https://native.hotwired.dev/reference/navigation#server-driven-routing-in-rails):
```ruby
recede_or_redirect_to(url, **options) # Pops the visible screen off of the navigation stack. If a modal is presented on iOS, the modal is dismissed instead.
resume_or_redirect_to(url **options) # No action is taken.
refresh_or_redirect_to(url, **options) # Reloads the visible screen by performing a new web request and invalidating the cache.
```
which I found necessary to get forms and modals working properly with the template. @joemasilotti would you be open to a PR adding the routes? e.g.
```json
        {
            "patterns": [
                "/recede_historical_location"
            ],
            "properties": {
                "presentation": "pop"
            }
        },
        {
            "patterns": [
                "/resume_historical_location"
            ],
            "properties": {
                "presentation": "none"
            }
        },
        {
            "patterns": [
                "/refresh_historical_location"
            ],
            "properties": {
                "presentation": "refresh"
            }
        }
```

joemasilotti:
This is something we'd love to see included by default! Ideally, this would be built in to the codebase and not require any path configuration to work. So I'm not 100% sure of the implementation path, yet. But if you can get the ball rolling there's a good chance it will help us get it over the finish line.

jpsphaxer:
Hey @joemasilotti - I would love to get some high-level feedback on the draft PR I opened when you get a chance.

subelsky:
@jpsphaxer that's cooler than what I was thinking :-)

jpsphaxer:
> [@jpsphaxer](https://github.com/jpsphaxer) that's cooler than what I was thinking :-)

Thanks! Another approach I thought about was including those `PathRules` as default and merging them with whatever rules we get from a file/server.  But I am open to suggestions!

svara:
> > [@jpsphaxer](https://github.com/jpsphaxer) that's cooler than what I was thinking :-)
> 
> Thanks! Another approach I thought about was including those `PathRules` as default and merging them with whatever rules we get from a file/server. But I am open to suggestions!

@jpsphaxer This is what I had in mind too. The advantage of this approach is that we don't pollute the `NavigationHierarchyController ` and introduce additional complexity.
Have a look at https://github.com/hotwired/hotwire-native-ios/pull/76.

cc @jayohms @joemasilotti

[...]

PR #74, opened 2025-01-23
server driven navigation without path configuration

jpsphaxer:
This PR addresses [Issue #71](https://github.com/hotwired/hotwire-native-ios/issues/71), which proposed adding support for server-driven routes like `/recede_historical_location`, `/resume_historical_location`, and `/refresh_historical_location`. These routes are designed to handle navigation stack behaviors (pop, refresh, or no-op) based on directives from the server.

## Implementation Overview
I've introduced a new extension to the NavigationHierarchyController class to handle these historical location routes. Key changes include:

### Route Handling:

- Added a handleHistoricalLocation(proposal:) method to process specific historical routes and execute actions such as:
  - `recede_historical_location`: Pops the current view or dismisses a modal.
  - `resume_historical_location`: No-op (does nothing).
  - `refresh_historical_location`: Refreshes the current view via a VisitProposal.

### Route Detection:

Added an `isHistoricalLocation(_:)` helper method to determine if a given URL matches one of the `_historical_location` patterns.
### Routing Logic:

Updated `route(controller:proposal:)` to check for historical location routes before falling back to default routing logic.

## Goals of This PR

- **Start the Discussion:** This is a proof of concept, and I’d like to gather feedback on whether this is the right direction for implementation.
- **Evaluate the Approach:** Does adding this logic to NavigationHierarchyController align with the desired architecture? Is there a better way to integrate this functionality natively without requiring explicit path configurations?
- **Explore Next Steps:** If this approach works, we could look into automating or simplifying the configuration process.

### Notes: 
I will be adding tests in the coming days - just wanted to see what the initial reaction might be.

### DIFF
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -2,6 +2,32 @@ import SafariServices
 import UIKit
 import WebKit
 
+extension NavigationHierarchyController {
+    /// Handles special Turbo historical location routes
+       private func handleHistoricalLocation(proposal: VisitProposal) {
+           switch proposal.url.path {
+           case "/recede_historical_location":
+               // Pop or dismiss the current view
+               pop(animated: proposal.animated)
+
+           case "/resume_historical_location":
+               // No-op - do nothing
+               break
+               
+           case "/refresh_historical_location":
+               // refresh current view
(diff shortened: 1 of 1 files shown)


jpsphaxer:
Closing this draft in favor of [PR#76](https://github.com/hotwired/hotwire-native-ios/pull/76)

PR #75, opened 2025-01-24
Expose userAgent as get only property

jpsphaxer:
# Expose User Agent String in HotwireConfig

## Motivation
Previously in TurboNavigator, we could access the user agent string through TurboConfig. This functionality was lost in the migration to HotwireNative. Applications often need to maintain consistent user agent strings across different components (HTTP clients, web views, etc.) to ensure proper tracking and identification.

## Changes
- Added read-only `userAgent` property to HotwireConfig
- Added test to verify property functionality

## Testing
Added `HotwireConfigTests` with coverage for user agent string generation and compile-time verification of read-only access.

No breaking changes were introduced.

### DIFF
--- a/Source/HotwireConfig.swift
+++ b/Source/HotwireConfig.swift
@@ -29,6 +29,17 @@ public struct HotwireConfig {
             HotwireLogger.debugLoggingEnabled = debugLoggingEnabled
         }
     }
+    
+    /// Getter for user-agent string
+    /// - returns: "`applicationUserAgentPrefix`; Native iOS; Turbo Native iOS; bridge-components: [your bridge components];"
+    public var userAgent: String {
+        get {
+            return UserAgent.build(
+                applicationPrefix: applicationUserAgentPrefix,
+                componentTypes: Hotwire.bridgeComponentTypes
+            )
+        }
+    }
 
     // MARK: Turbo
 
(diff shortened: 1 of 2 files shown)


joemasilotti:
Thanks for this @jpsphaxer! I'm all for making this property get-able, too.

My only concern is the name. `userAgent` might imply the full user agent of the application, with WebKit and all that. But that isn't the case - we actually can't read that from `WKWebView` at all!

I'm thinking we might want to rename the getter to something that makes it obvious this is only the part provided by Hotwire Native and not the system. I'm open to ideas on what a good name might be.

jpsphaxer:
> Thanks for this @jpsphaxer! I'm all for making this property get-able, too.
> 
> My only concern is the name. `userAgent` might imply the full user agent of the application, with WebKit and all that. But that isn't the case - we actually can't read that from `WKWebView` at all!
> 
> I'm thinking we might want to rename the getter to something that makes it obvious this is only the part provided by Hotwire Native and not the system. I'm open to ideas on what a good name might be.

what do you think of `hotwireUserAgentComponent` ?  - naming is never my strong suit 😅

joemasilotti:
Maybe even just `userAgentComponent` since we are already in the Hotwire Native space.

What do you think @svara?

jpsphaxer:
@joemasilotti - went ahead and updated the naming to `userAgentComponent` for the time being unless we decide to change the naming.

svara:
> Maybe even just `userAgentComponent` since we are already in the Hotwire Native space.
> 
> What do you think @svara?

Sorry for late reply. `userAgentComponent` sounds good to me 👍🏻

svara (approved):
Nicely done @jpsphaxer 👏🏻

svara:
FYI, we've decided to rename `userAgentComponent` to `userAgent` to align the API available in the Android library https://github.com/hotwired/hotwire-native-android/pull/94.

PR #76, opened 2025-01-24
Include default server route rules by default

svara:
This PR builds on top of https://github.com/hotwired/hotwire-native-ios/pull/74 and addresses https://github.com/hotwired/hotwire-native-ios/issues/71.

### Context
[Turbo Native](https://native.hotwired.dev/reference/navigation#server-driven-routing-in-rails) allows server-driven routing in Rails applications. Previously, developers had to explicitly define these routing rules in the path config. This PR adds default routing rules for server-driven navigation directly into the framework.

### Changes
- recede, resume and refresh historical location rules (historical location rules) have been added as `PathRule`s.
- historical location rules are now loaded by default whether or not the source of the path config is provided.
- the `NavigationHierarchyController` has been adapted to handle the historical location rules.

### Historical Location Behaviour

**resume_historical_location**
1. Dismiss the modal view controller.
2. Arrive back at the view controller on the "default" stack.

**refresh_historical_location**
1. Dismiss the modal view controller.
2. Arrive back at the view controller on the "default" stack.
3. Refresh the view controller no the "default" stack by revisiting the location.

**recede_historical_location**
1. Dismiss the modal view controller.
2. Arrive back at the view controller on the "default" stack.
3. Pop the view controller on the "default" stack (unless it's already at the beginning of the backstack).
4. This will trigger a refresh on the appeared view controller if the snapshot cache has been cleared by a form submission.

### DIFF
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -63,7 +63,7 @@ class NavigationHierarchyController {
     }
 
     func pop(animated: Bool) {
-        if navigationController.presentedViewController != nil {
+        if navigationController.presentedViewController != nil && !modalNavigationController.isBeingDismissed {
             if modalNavigationController.viewControllers.count == 1 {
                 navigationController.dismiss(animated: animated)
             } else {
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -41,6 +41,8 @@ class NavigationHierarchyController {
                 visitable.visitableView.allowsPullToRefresh = proposal.pullToRefreshEnabled
             }
 
+            dismissModalIfNeeded(for: proposal)
+
             switch proposal.presentation {
             case .default:
                 navigate(with: controller, via: proposal)
@@ -175,6 +177,11 @@ class NavigationHierarchyController {
     }
 
     private func refresh(via proposal: VisitProposal) {
+        if proposal.isHistoricalLocation {
+            refreshIfTopViewControllerIsVisitable(from: .main)
+            return
+        }
+
         if navigationController.presentedViewController != nil {
             if modalNavigationController.viewControllers.count == 1 {
                 navigationController.dismiss(animated: proposal.animated)
@@ -183,10 +190,11 @@ class NavigationHierarchyController {
                 modalNavigationController.popViewController(animated: proposal.animated)
                 refreshIfTopViewControllerIsVisitable(from: .modal)
             }
-        } else {
-            navigationController.popViewController(animated: proposal.animated)
-            refreshIfTopViewControllerIsVisitable(from: .main)
+            return
         }
+
+        navigationController.popViewController(animated: proposal.animated)
+        refreshIfTopViewControllerIsVisitable(from: .main)
     }
 
     private func replaceRoot(with controller: UIViewController, via proposal: VisitProposal) {
@@ -203,5 +211,16 @@ class NavigationHierarchyController {
             delegate.refreshVisitable(navigationStack: stack,
                                                   newTopmostVisitable: navControllerTopmostVisitable)
         }
+    }
+
+    private func dismissModalIfNeeded(for visit: VisitProposal) {
+        // The desired behaviour for historical location visits is
+        // to always dismiss the "modal" stack.
+        guard visit.isHistoricalLocation,
+              navigationController.presentedViewController != nil else {
+            return
+        }
+
+        navigationController.dismiss(animated: visit.animated)
     }
 }
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -2,6 +2,32 @@ import SafariServices
 import UIKit
 import WebKit
 
+extension NavigationHierarchyController {
+    /// Handles special Turbo historical location routes
+       private func handleHistoricalLocation(proposal: VisitProposal) {
+           switch proposal.url.path {
+           case "/recede_historical_location":
+               // Pop or dismiss the current view
+               pop(animated: proposal.animated)
+
+           case "/resume_historical_location":
+               // No-op - do nothing
+               break
+               
+           case "/refresh_historical_location":
+               // refresh current view
+               refresh(via: proposal)
+           default:
+               break
+           }
+       }
+       
+       /// Check if URL is a Turbo historical location directive
+       private func isHistoricalLocation(_ url: URL) -> Bool {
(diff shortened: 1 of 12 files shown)


jpsphaxer:
I like this approach @svara - thanks for sharing this!

Tests/Turbo/Navigator/NavigationHierarchyControllerHistoricalLocationTests.swift
```diff
@@ -0,0 +1,148 @@
...
+        try await delay()
+
+        XCTAssertEqual(modalNavigationController.viewControllers.count, 1)
+
+        let recedeHistoricalLocationProposal = VisitProposal(
+            path: PathRule.recedeHistoricalLocation.patterns.first!,
+            additionalProperties: PathRule.recedeHistoricalLocation.properties
+        )
+        navigator.route(recedeHistoricalLocationProposal)
+        try await delay()
```
svara, line 121: The delays are far from ideal, but they are necessary to ensure the view lifecycle methods execute properly and the tests pass. I’ll explore architectural improvements in the future.

Source/Turbo/Path Configuration/PathRule+ServerRoutes.swift
```diff
@@ -0,0 +1,36 @@
...
+        .resumeHistoricalLocation,
+        .refreshHistoricalLocation
+    ]
+
+    static let recedeHistoricalLocation = PathRule(
+        patterns: ["/recede_historical_location"],
+        properties: [
+            "presentation": "pop",
+            "visitable": false,
+            "historical_location": true
```
jayohms, line 15: It's possible that someone includes these historical location paths in their files, attempting to override the default behavior. Are there other properties that we should explicitly provide here? I'm thinking we should at least provide `"context": "default"` for each of these rules to make it clear these don't operate in the modal context.

Source/Turbo/Path Configuration/PathRule+ServerRoutes.swift
```diff
@@ -0,0 +1,36 @@
...
+        .recedeHistoricalLocation,
+        .resumeHistoricalLocation,
+        .refreshHistoricalLocation
+    ]
+
+    static let recedeHistoricalLocation = PathRule(
+        patterns: ["/recede_historical_location"],
+        properties: [
+            "presentation": "pop",
+            "visitable": false,
```
jayohms, line 14: Is `visitable` already being used in the library? If so, how/where is it being used? Android doesn't have a corresponding property, so I'm curious if there's anything to align here.

jayohms (commented):
Great work on this @svara. Just a couple questions to start.

Source/Turbo/Navigator/NavigationHierarchyController.swift
```diff
@@ -204,4 +212,15 @@ class NavigationHierarchyController {
...
+
+    private func dismissModalIfNeeded(for visit: VisitProposal) {
+        // The desired behaviour for historical location visits is
+        // to always dismiss the "modal" stack.
+        guard visit.isHistoricalLocation,
+              navigationController.presentedViewController != nil else {
+            return
+        }
+
+        navigationController.dismiss(animated: visit.animated)
```
joemasilotti, line 224: The guard with the `!= nil` feels like a lot of inverse logic. What if we used an `if` statement here instead?

```swift
if visit.isHistoricalLocation, navigationController.presentedViewController != nil {
    navigationController.dismiss(animated: visit.animated)
}
```

Tests/Turbo/Navigator/NavigationHierarchyControllerHistoricalLocationTests.swift
joemasilotti: I'm surprised these are needed even with the `TestableNavigationController` presenting everything without animation. Does pushing, popping, or dismissing require the delay?

Source/Turbo/Path Configuration/PathRule+ServerRoutes.swift
svara: Good point! They could, but since the default rules are added last, they will always override those in the path config. I'll add `"context": "default"` and also a test.

Source/Turbo/Path Configuration/PathRule+ServerRoutes.swift
svara: Good catch! It isn't. We use it internally in the email app. I'll remove it.

Source/Turbo/Navigator/NavigationHierarchyController.swift
svara: I personally don't have issues with this and use `guard` extensively for early returns. If you feel strongly about this we can switch to `if`.

Tests/Turbo/Navigator/NavigationHierarchyControllerHistoricalLocationTests.swift
svara: Without the delay the `recede historical location` test fails. Even though the presentations are not animated, I doubt the view lifecycle methods are all called synchronously. I'll investigate.

Source/Turbo/Navigator/NavigationHierarchyController.swift
svara: @joemasilotti I've refactored the code. Let me know if it's easier to understand, otherwise I'll go with the `if` statement.

joemasilotti: Oh, I like this! Thank you.

Tests/Turbo/Navigator/NavigationHierarchyControllerHistoricalLocationTests.swift
klaaspieter: I got nerd-sniped and did some digging and while I still can't fully explain the need for `delay` I am surprised by the recede behavior.

I implemented a test project that follows the same steps as this test: 1) push `default_one` 2) push `default_two` 3) present `modal_one` and then call `popViewController` on the original navigation controller.

What happens is that `modal_one` remains visible while `default_two` is being popped in the background. I would expect `recede` to dismiss the presented view controller and only if there isn't one pop the navigation stack.

As for the delay _something_ appears to be happening on the next runloop. Which means the delay may be necessary regardless but a delay of 0 seconds is enough because that allows the current runloop to finish.

svara: @klaaspieter I tweaked the tests and was able to remove the delay. 
As for the recede behaviour, your implementation doesn't reproduce it correctly. See step 4.
1) push default_one 
2) push default_two 
3) present modal_one 
4) ~call popViewController on the original navigation controller.~ dismiss modal, and then pop the view controller on the main navigation controller

Issue #48, opened 2024-11-22
Turbo frame action "advance" is ignored


nunesmatheus:
Hey, thanks for this amazing project. It really makes me excited about web development again!

## Bug

As I was using it to build my first iOS application, I noticed a bug that causes my application to work like this:
1. Go to a page that supports pagination
2. Click a pagination link to go the next page (page 2)
3. Refresh the page
4. **Result**: the refreshed page is page 1
5. **Expected result**: the refreshed page is page 2

I think this happens because I'm using \<turbo-frame\> with `data-turbo-action` attribute as "advance" and it is not being respected, so the navigation works fine, but the URL doesn't change and a refresh sends us back to the initial page that rendered the original turbo-frame.
Basically, I'm expecting this to work natively as well, so correct me if I'm wrong on that assumption: https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit.

## Steps to reproduce

1. Run an app that has a turbo-frame with navigation promoted to page visit
2. Open the demo app from this repo on Xcode, point to the application from above and run it
3. Navigate to a page that has pagination
4. Go to the next page
5. Pull to refresh the page and see it go back to the previous page instead of refreshing the current page

I created this tiny Rails app that can be used to test it: https://github.com/nunesmatheus/turbo-frame-promote-visit-demo.

I just ran `rails new` and added a simple page with pagination links inside a turbo frame that promotes navigation to page visits: https://github.com/nunesmatheus/turbo-frame-promote-visit-demo/commit/950dcd9c15b359af4cb025b7372f13aaf03eecc4.

ibrahima:
I believe I'm seeing this as well. We just started using turbo frames and now the navigation in our Hotwire Native app is basically frozen when you hit a path that uses turbo frames.

leonvogt:
Hey there,
No guarantee, but I think PR #77 could solve this issue

ibrahima:
In the issue body you primarily focused on the issue that reloading is broken currently. Based on [the documentation you linked](https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit), it seems like it _should_ also add to the navigation history, but it doesn't right now. #77 solves the reload issue but does not affect the navigation history.

[...]

PR #77, opened 2025-01-26
Update the `visitableURL` on reload

leonvogt:
Hi there,

Sorry for the spam lately 😅
While working on a feature I've noticed that Hotwire Native iOS handles the pull to refresh feature slightly different than Android.   

We use HTML tabs on some views and a bit of JavaScript to update the URL whenever a user switches tabs.   
This ensures that when the page is refreshed, the correct tab is selected.
The JavaScript code looks something like this:
```javascript
tab.addEventListener('click', () => {
  const currentURL = new URL(window.location);
  currentURL.searchParams.set('view_part', tab.dataset.bsTarget)
  window.history.replaceState(history.state, "", currentURL)
})
```

Here's the current behavior:

**Android**

https://github.com/user-attachments/assets/06a28fec-4c1c-4dd8-a13e-3dd4c1382563

**iOS**

https://github.com/user-attachments/assets/bff09688-ae4d-4146-a5a3-8c8df64df763

---

On iOS, when a refresh is triggered, it reloads the (visitable)URL that was set at the start of the ViewController.
This PR updates iOS to behave like Android by ensuring it reloads the current URL from the webview instead.

### DIFF
--- a/Source/Turbo/Session/Session.swift
+++ b/Source/Turbo/Session/Session.swift
@@ -85,7 +85,10 @@ public class Session: NSObject {
 
     public func reload() {
         guard let visitable = topmostVisitable else { return }
-
+        if let currentURL = webView.url {
+            visitable.visitableURL = currentURL
+        }
+        
         initialized = false
         visit(visitable)
         topmostVisit = currentVisit
--- a/Source/Turbo/Visitable/Visitable.swift
+++ b/Source/Turbo/Visitable/Visitable.swift
@@ -14,7 +14,7 @@ public protocol Visitable: AnyObject {
     var visitableViewController: UIViewController { get }
     var visitableDelegate: VisitableDelegate? { get set }
     var visitableView: VisitableView! { get }
-    var visitableURL: URL! { get }
+    var visitableURL: URL! { get set }
 
     func visitableDidRender()
     func showVisitableActivityIndicator()


ibrahima:
Hi @leonvogt! Saw your comment on #48, I was trying to see if it would resolve my issue. But I am not sure how to use your fork of the library in my Xcode project, do you have any suggestions? I tried overriding the revision in `Package.resolved` but I don't know if it works that way. I set it to this and it didn't seem to resolve my issue but I don't know whether it's actually using that version or not.

```json
		{
			"identity": "hotwire-native-ios",
			"kind": "remoteSourceControl",
			"location": "https://github.com/leonvogt/hotwire-native-ios",
			"state": {
				"revision": "272249c5c825338a61f6fa4dcd1429a7e1fde789",
				"version": "1.1.1"
			}
		},
```

(I am also building the app through fastlane as I don't have access to my Mac computer right now... so that's also a confounding factor.)

leonvogt:
Hey @ibrahima,
Happy to hear that you figured it out and it's working for you!   

> However, it doesn't add an entry to the navigation stack even though I have data-turbo-action="advance" set, so that's still an issue IMO

Yeah, I guess that's a separate issue. The native Turbo adapter would need to intercept the request and propagate it as a ordinary request. So you get a new ViewController on the stack. But I guess that would open a whole new set of problems, since your server may only response with the new Turbo Frame content and not with a whole page.    
Also, you maybe don't always want that kind of behaviour. In the example of #48 with pagination, it would be kind of weird to push something on the stack, every time you click on a pagination link.

joemasilotti:
This looks good to me. But I'd love to get a second (or third!) set of eyes on it to make sure it doesn't have unintended consequences. @svara, what do you think?

svara:
@joemasilotti We had a similar issue in Strada -> https://github.com/hotwired/strada-ios/pull/24.
@dginsburg provided a thorough [explanation](https://github.com/hotwired/strada-ios/pull/24#issuecomment-1981852025) and I agree with him. Let me work on some feedback.

joemasilotti:
Nice digging, I had almost forgotten about that conversation. Thanks @svara, I'll wait on you.

svara:
Given that the location of a page can change after the view controller is created, I think we should handle this in `VisitableViewController`. We can leverage `visitableView` to retrieve the current URL directly from the web view.
However, this may introduce navigation issues, as `NavigationHierarchyController` relies on the `visitableURL` for some of its logic.
This change has broader implications, so I’d like to discuss it with @jayohms.

svara:
Follow-up work in https://github.com/hotwired/hotwire-native-ios/pull/108.
@leonvogt Would you mind giving it a test please?

leonvogt:
Just tested it—works like a charm, @svara! 
Thank you very much 🙏

svara:
Amazing! Thanks @leonvogt and sorry for the delay.

svara:
Closing in favor of https://github.com/hotwired/hotwire-native-ios/pull/108.

PR #79, opened 2025-01-30
add `userAgentWithWebViewDefault`on iOS

jpsphaxer:
# Expose and Align User Agent String Handling w/ Hotwire Native Android

## Motivation
Following the Android implementation (hotwired/hotwire-native-android#94), we need proper user agent string handling in iOS to ensure consistent user agent strings across WebViews and provide access to both the custom Hotwire user agent and the full WebKit user agent string.

## Changes
- Made `cachedUserAgent` internal to allow for proper testing and cache management
- Added documentation for user agent properties
- Added tests to verify user agent string generation and caching behavior
- Ensures tests properly clean up static state between runs to prevent test pollution

## Implementation Details
We're adopting a similar approach to the Android implementation where we:
1. Cache the WebKit default user agent string when a WebView is first created
2. Provide access to both:
  - The custom Hotwire user agent parts (`userAgent`)
  - The full user agent including WebKit's default (`userAgentWithWebViewDefault`)

## Testing
Added tests to verify:
- Custom user agent string generation
- Full user agent string composition
- Proper cache management between test runs

The changes ensure consistent behavior across test runs by properly cleaning up static state.

### DIFF
--- a/Source/HotwireConfig.swift
+++ b/Source/HotwireConfig.swift
@@ -45,6 +45,20 @@ public struct HotwireConfig {
             )
         }
     }
+    
+    /// Gets the full user agent that is used for every WKWebView instance. This includes:
+    /// - The WKWebView's default WebKit user agent string
+    /// - Your (optional) custom `applicationUserAgentPrefix`
+    /// - "Hotwire Native iOS; Turbo Native iOS;"
+    /// - "bridge-components: [your bridge components];"
+    public var userAgentWithWebViewDefault: String {
+        get {
+            if let defaultAgent = Self.cachedUserAgent {
+                return "\(defaultAgent)"
+            }
+            return userAgent
+        }
+    }
 
     // MARK: Turbo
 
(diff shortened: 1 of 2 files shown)


jpsphaxer:
@joemasilotti @svara @jayohms - wanted to see what your thoughts are towards exposing the "full" user agent string on iOS so that we can align with Hotwire Native on Android? - is this something we would like to explore and implement?

joemasilotti:
As much as I'd like to have feature parity with Android, this is *technically* using a private API: `webView.value(forKey: "userAgent")`.

If we really need to expose this then we should probably be using JavaScript to grab `navigator.userAgent`. But that requires an async operation and I'm not sure how useful that would be for developers.

In my experience, I've never needed the web view's default user agent. If I'm making a native request I'd rather set something custom and append the Hotwire Native stuff to the end. Do you have a use-case for needing this?

I don't personally think we need to add this to the iOS library.

jpsphaxer:
@joemasilotti - that’s a fair point.  If we are gonna use JS we can devs could just use a bridge component also.

jayohms:
Thanks for understanding. A main reason that I've exposed this on the Android side is because the Chromium `WebView` component version is not tied to the OS at all. It's updated via the Play Store, so it's possible to have a user on a newer OS, with an outdated `WebView` component installed. At 37signals, we attach the full (WebView included) user agent string to support requests so that we have an understanding of the OS/WebView versions are installed on the device.

Outside of that, say for native API requests, the (now) public `Hotwire.config.userAgent` should be sufficient for both platforms. `WKWebView` on iOS is tied to the OS version, so there's not the same ambiguity that exists on Android.

Issue #78, opened 2025-01-28
Cross-origin redirect inside of an iframe is interpreted as a redirect for the app


glennfu:
I have a page loading with a standard inline YouTube embed:
```
<iframe
  width="480"
  height="270"
  src="https://www.youtube.com/embed/dQw4w9WgXcQ?playsinline=1&showinfo=0&rel=0&iv_load_policy=3&color=white"
  frameborder="0"
  allowfullscreen
></iframe>
```
Since updating from 1.1.0 to 1.1.1, I assume as a result of #70, when this page loads it instantly bounces over to the Safari app at this YouTube url.

I'm not sure what the best solution is, but my hope would be that in detecting things like Cross-origin redirects, it could safely ignore those that aren't happening on the parent frame.

To reproduce, you can take the Demo app and locally serve up just the iframe html above.

noah44846:
I also encountered the same issue, but for the Google Maps embed. For me, it also worked in 1.1.0 but not in 1.1.1. Here is the iframe:

```html
<iframe
    allowfullscreen=""
    class="mb-2 rounded-3"
    frameborder="0"
    height="150"
    width="100%"
    src="https://www.google.com/maps/embed/v1/place?key=<key...>&q=<query...>&zoom=15"
></iframe>
```

The thing is that for me when I navigate to the page the containing the iframe there is no issue, but when I pull-to-refresh the WebView it opens the embedded link in the browser.

darthdie:
We're encountering the same thing and have opened up a reproducible example here:
[hotwired/hotwire-native-demo#13](https://github.com/hotwired/hotwire-native-demo/pull/13) 

It seems to only happen during a ColdBootVisit for us.

Adding

```
        guard let targetFrame = navigationAction.targetFrame, targetFrame.isMainFrame else {
            decisionHandler(.cancel)
            return
        }
```

into `ColdBootVisit func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void)`

Fixes the issue.

svara:
Hey guys, we've merged https://github.com/hotwired/hotwire-native-ios/pull/80 which should address this issue.
Could you give it a test on main? Thanks!

> The thing is that for me when I navigate to the page the containing the iframe there is no issue, but when I pull-to-refresh the WebView it opens the embedded link in the browser.

[...]

PR #80, opened 2025-01-30
Ignore cross-origin redirects that aren’t happening in the main frame

svara:
This PR addresses https://github.com/hotwired/hotwire-native-ios/issues/78 by ensuring that only top-level navigations are subject to cross-origin redirect checks, while allowing iframe and sub-frame navigations to proceed without interference.

This prevents unwanted redirects from embedded content (e.g., YouTube iframes) while maintaining normal navigation behaviour in the parent frame.

### DIFF
--- a/Source/Turbo/Visit/ColdBootVisit.swift
+++ b/Source/Turbo/Visit/ColdBootVisit.swift
@@ -75,6 +75,13 @@ extension ColdBootVisit: WKNavigationDelegate {
             return
         }
 
+        // Allow navigation if it happens inside an iframe (sub-frame).
+        // This prevents Turbo Native from mistakenly treating iframe loads as full navigations.
+        guard let targetFrame = navigationAction.targetFrame, targetFrame.isMainFrame else {
+            decisionHandler(.allow)
+            return
+        }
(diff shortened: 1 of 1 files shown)


Source/Turbo/Visit/ColdBootVisit.swift
```diff
@@ -75,6 +75,13 @@ extension ColdBootVisit: WKNavigationDelegate {
             return
         }
 
+        // Allow navigation if it happens inside an iframe (sub-frame).
+        // This prevents Turbo Native from mistakenly treating iframe loads as full navigations.
```
jayohms, line 79: Instead of "Turbo Native" (which we're not using anymore), how about something like:

> This prevents iframe loads from being treated as a full navigation during cold boots.

jayohms (approved):
Nice, thanks!

Source/Turbo/Visit/ColdBootVisit.swift
svara: "Turbo Native" 🙈 
I've updated the comment as suggested.

Issue #83, opened 2025-02-14
bridgeDidIgnoreMessage triggered after FormSubmission on modal


williamkennedy:
## Intro

Great work on hotwire-native-ios. The evolution from turbo-ios has been a joy to work with. 

I'm unsure if this problem is a stimulus or a hotwire-native-ios issue. I have not tried to replicate this on Android just yet.

## Behavior

When submitting a form with success and redirect, the bridge component raises the error message `bridgeDidIgnoreMessage`. It seems that the message is rejected and called from this [line](https://github.com/hotwired/hotwire-native-ios/blob/0bc9f80d1357db3a46a30578b574ded30a23e2ec/Source/Bridge/BridgeDelegate.swift#L129)

This means the bridge does not render on the show page after redirect.

## Expected Behavior 

The bridge component should be rendered. 

## Video

https://share.cleanshot.com/L16qxynS

## Source code

Branch is [here](https://github.com/williamkennedy/hotwire-native-demo/tree/bridgeDidIgnoreMessage) and PR to show file change is [here](https://github.com/hotwired/hotwire-native-demo/pull/14/files)

## Logs

```
[Session] visit ["reload": false, "location": http://localhost:45678, "options": HotwireNative.VisitOptions(action: HotwireNative.VisitAction.advance, response: nil)]
[ColdBootVisit] startVisit http://localhost:45678
[Bridge] bridgeDestinationViewDidLoad: http://localhost:45678/
[Bridge] bridgeDestinationViewWillAppear: http://localhost:45678/
Failed to send CA Event for app launch measurements for ca_event_type: 0 event_name: com.apple.app_launch_measurement.FirstFramePresentationMetric
[Bridge] bridgeDestinationViewDidAppear: http://localhost:45678/
Failed to send CA Event for app launch measurements for ca_event_type: 1 event_name: com.apple.app_launch_measurement.ExtendedLaunchMetrics
Failed to resolve host network app id to config: bundleID: com.apple.WebKit.Networking instance ID: Optional([_EXExtensionInstanceIdentifier: 0BAF1C8C-E894-4F1F-939E-9582A58E41FB])
[ColdBootVisit] Same-origin redirect detected: http://localhost:45678 -> http://localhost:45678/. http://localhost:45678
[...]

PR #84, opened 2025-02-14
Ensure destinationIsActive is set to true when webView is loaded

williamkennedy:
This is a possible fix for [issue 83](https://github.com/hotwired/hotwire-native-ios/issues/83)

### DIFF
--- a/Source/Bridge/BridgeDelegate.swift
+++ b/Source/Bridge/BridgeDelegate.swift
@@ -56,6 +56,7 @@ public final class BridgeDelegate: BridgingDelegate {
     public func webViewDidBecomeDeactivated() {
         bridge?.delegate = nil
         bridge = nil
+        destinationIsActive = false
     }
     
     @discardableResult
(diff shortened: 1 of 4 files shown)


joemasilotti:
This seems sensible to me. But I'll admit that I'm not very versed in the internals of the bridge library.

@svara, it looks you have the most recent changes via [this commit](https://github.com/hotwired/strada-ios/commit/cf2373428221193e242bf3b37fdb880c46501803). Any thoughts on if this could cause any unforeseen issues?

svara:
I don't foresee any issues. However, I'm wondering if Android is affected by this too? cc @jayohms

williamkennedy:
Hi @svara 

I was unable to recreate the issue on Android. 

CleanShot 2025-03-07 at 13 28 15

jayohms:
@svara I don't anticipate issues on the Android side due to the lifecycle differences, but happy to see a solution for iOS!

svara:
Thanks @williamkennedy and @leonvogt for giving it a test on Android.
Setting `destinationIsActive` to `true` when the web view becomes active makes sense. I took the liberty of balancing that by setting it to `false` when the web view becomes deactivated, and I added tests.

PR #87, opened 2025-02-18
Add tabs and use new Rails-powered demo site

joemasilotti:
This PR works with the new Ruby on Rails demo site and adds a native tab bar, [PR #16](https://github.com/hotwired/hotwire-native-demo/pull/16).

The only Source change is moving the Done button on modals from the left side to the right side. This is to ensure the Done button doesn't override the back button on a second modal screen.

The tab implementation is quite naive - as soon as the app launches all three tabs are loaded at the same time.

The appearance of the navigation and tab bars are also set to not be transparent. This fixes an issue where the web content goes *under* a native bar when over scrolling.

This PR is now branched off of @svara's work in #76.


svara:
This is a great start @joemasilotti! 
I think we can take this a step further and provide a Hotwire tab view controller. I've started working on it in https://github.com/hotwired/hotwire-native-ios/pull/109.

Demo/SceneController.swift
```diff
@@ -25,26 +22,25 @@ extension SceneController: UIWindowSceneDelegate {
...
         case NumbersViewController.pathConfigurationIdentifier:
-            return .acceptCustom(NumbersViewController(url: proposal.url, navigator: navigator))
-
-        case "numbers_detail":
-            let alertController = UIAlertController(title: "Number", message: "\(proposal.url.lastPathComponent)", preferredStyle: .alert)
-            alertController.addAction(.init(title: "OK", style: .default, handler: nil))
-            return .acceptCustom(alertController)
+            return .acceptCustom(NumbersViewController(
+                url: proposal.url,
+                navigator: tabBarController.activeNavigator
```
joemasilotti, line 38: @svara, what's our recommendation here? The active navigator or the one that triggered the request?

jayohms: I think the default recommendation should be the navigator that triggered the request.

svara: Agree.

svara (approved):
Looks great @joemasilotti!

Issue #89, opened 2025-02-20
Figuring out if a BridgeComponent is in a modal


tim-vandecasteele:
I want to write functionality for a BridgeComponent that acts differently if you're running in a modal or not. But I have a very hard time figuring out if the component is in a model. Initially I went through checking the `activeNavigationController`, and checking if it has a `presentingViewController`, but the problem here is that if the page loads quick enough, and the modal is still dismissing, that this seems to trigger false positives.

It looks rather simple if you would just check to what `Session` the `Webview` belongs. I could easily compare those, but it looks like everything in the Navigator is done to make sure you can't access the two `Session`s. The `init` where you can set the different `Session`s is internal, and both the `modalSession` and the `session` are internal as wel.

Is there a good reason for those to be internal? Would you be ok exposing this in some way, so I could access them for comparison?

joemasilotti:
Closed via #90.

PR #90, opened 2025-02-21
Make session and modalSession public

tim-vandecasteele:
As discussed in #89 I would want to check if a webview belongs to the session or the modalSession. This PR makes these public in the navigator, so I can access them with an extension. (I haven't added any additional code as I didn't know if this would be wanted)

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -19,6 +19,8 @@ public class Navigator {
         }
         return modalSession.webView
     }
+    public var session: Session
+    public var modalSession: Session
     
     /// Set to handle customize behavior of the `WKUIDelegate`.
     ///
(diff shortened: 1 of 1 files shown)


jayohms:
@joemasilotti I don't have any issue exposing these, do you? It'd give flexibility to developers without having to explicitly add new features ourselves. These are already exposed through `navigator.session` on the Android side.

Source/Turbo/Navigator/Navigator.swift
```diff
@@ -19,6 +19,8 @@ public class Navigator {
         }
         return modalSession.webView
     }
+    public var session: Session
+    public var modalSession: Session
```
joemasilotti, line 23:
```suggestion
    public private(set) var session: Session
    public private(set) var modalSession: Session
```

joemasilotti:
If these are already exposed in Android then yes, let's also expose them in iOS.

But they probably shouldn't be able to be changed outside of `Navigator`. I made a suggestion in the PR.

jayohms:
Ok perfect. And yes, of course, these should only be public getters 👍

tim-vandecasteele:
Thanks, this solution sounds perfect! Having the setters private is indeed super logical. This would definitely solve my problem!

Issue #88, opened 2025-02-20
Users report frozen screens


michaelloistl:
After some investigation, it turned out that the screenshot is being shown which creates for the user the impression that the screen is frozen. I have not been able to recreate this issue myself, but got countless reports from users. I distributed a TestFlight build with color coded screen borders for all the functions that call `showVisitableScreenshot()`. Based on the color reported back from users, the following function would cause that frozen screen for them:

```swift
private func deactivateVisitable(_ visitable: Visitable, showScreenshot: Bool = false) {
  guard isActivatedVisitable(visitable) else { return }

  if showScreenshot {
    visitable.updateVisitableScreenshot()
    visitable.showVisitableScreenshot()

    // Color red added for test purposes
    visitable.visitableView.layer.borderWidth = 5
    visitable.visitableView.layer.borderColor = UIColor.red.cgColor
  }

  visitable.deactivateVisitableWebView()
  activatedVisitable = nil
}
```

The following screen recording shows when that happens to the user. We have however received reports that this happens across all of the tabs, so it's not unique to that scenario where a user comments on a post and returns to the feed.

https://github.com/user-attachments/assets/74a96aab-0caf-46c7-a8b7-797f7ac2fa82

joemasilotti:
Thanks for reporting. After some poking around I can reproduce with the [new version of the demo server](https://github.com/hotwired/hotwire-native-demo/pull/16) that [uses tabs](#87) on iOS.

https://github.com/user-attachments/assets/6162f0fb-f48f-46cf-bef8-279b04112afc

The freeze can be reproduced by:

1. Submitting a form
2. Switching to a different tab
3. Switching back to the original tab
4. Tapping the tab button (again) to pop to root

Not 100% of the time, though. So I'm wondering if there is a race condition occurring somewhere.

If I had to guess, this might be related to #50 by @svara. Only because that's the only custom code we have in place around tabs and such. There's also a high possibility the way I'm setting up tabs in #87 is causing the issue.

michaelloistl:
@joemasilotti as this is really urgent (having frustrated users & client), has there been any progress in fixing this issue or a suggestion for a hotfix?

joemasilotti:
Digging in some more, it looks like the `guard` statement in the code you mentioned is incorrectly returning `false` when switching between tabs (sometimes). I'll post more information as I uncover it.

```swift
private func deactivateVisitable(_ visitable: Visitable, showScreenshot: Bool = false) {
[...]

PR #91, opened 2025-02-27
Set appear reason when a modal is dismissed

joemasilotti:
This PR hopes to address #88. See that issue for reproduction steps.

The fix adds an additional hook to `HotwireNavigationController` that sets `VisitableViewController.appearReason` when a modal is dismissed.


svara (commented):
Nicely done @joemasilotti!
I went ahead and opened a feedback PR - https://github.com/hotwired/hotwire-native-ios/pull/99, where I also handle the modal presentation.

joemasilotti:
Closed in favor of #99.

Issue #94, opened 2025-03-02
Unexpected Modal refreshes


leonvogt:
This is a minor issue, but I thought I'd mention it anyway.

When inside a modal, a short pull-to-refresh gesture triggers a refresh of the modal content.   
At first, I thought I had forgotten to add the `pull_to_refresh_enabled` property to the modal configuration, but even with `pull_to_refresh_enabled: false`, the refresh still occurs.

https://github.com/user-attachments/assets/24932b98-f5ef-4f3b-9b10-06d1a694b0fe

- I couldn't reproduce this issue on Android, so it seems to be iOS-specific.
- It only occurs in the "medium" and "large" modal styles. I couldn't reproduce it in the "full" style.

This isn't very noticeable in most cases, but if the modal content differs significantly from the native background colors, the refresh effect becomes more visible and may be distracting.
It could interfere with form inputs by clearing the input fields when a refresh is triggered.

svara:
Closed via https://github.com/hotwired/hotwire-native-ios/pull/95.

PR #95, opened 2025-03-03
Fix unexpected modal refreshes

leonvogt:
This PR attempts to address #94.

The modal refresh occurred because the `viewWillAppear` method gets triggered even when the modal was already presented and only a minor height adjustment is made (via the short pull-to-refresh gesture).

There's nothing inherently wrong here, but the [context check for the view appearance](https://github.com/hotwired/hotwire-native-ios/blob/main/Source/Turbo/Session/Session.swift#L231-L282) misinterpreted this appearance as navigating backward from a native screen to a web view.   
With the pull-to-refresh gesture, we mistakenly entered [this if check inside the context check](https://github.com/leonvogt/hotwire-native-ios/blob/1ab960be5bd957284fda8bc2085e715a41c990e9/Source/Turbo/Session/Session.swift#L283-L286):
```swift
// Navigating backward from a native to a web view screen.
if visitable === previousVisit?.visitable {
    visit(visitable, action: .restore)
}
```

To fix this, I’ve added a simple guard clause to check if the topmost visitable is already the active visitable. This prevents the modal refresh from being triggered unnecessarily.

I’ve tested all the scenarios I could think of and verified that navigating from a native screen back to a web view still works as expected and displays the correct snapshot.
I haven’t noticed any unwanted side effects so far.

However, I still feel a bit unsure about this change and would highly appreciate it if someone could review this and let me know if I might be overlooking anything.

### DIFF
--- a/Source/Turbo/Session/Session.swift
+++ b/Source/Turbo/Session/Session.swift
@@ -274,6 +274,11 @@ extension Session: VisitableDelegate {
             visit(visitable, action: .restore)
             return
         }
+        
+        // If the topmost visitable is already the active visitable, nothing needs to be done
+        if topmostVisitable === activeVisitable {
+            return
+        }
 
         // Navigating backward from a native to a web view screen.
         if visitable === previousVisit?.visitable {


svara (approved):
Nicely done @leonvogt 👏🏻

PR #96, opened 2025-03-05
Complete navigation on restore

olivaresf:


joemasilotti:
@olivaresf, is this PR still relevant? I'm not sure which issue it resolving.

PR #98, opened 2025-03-05
Improved `default` context visit behavior when dismissing the `modal` context

olivaresf:
This PR fixes an issue when you're in the `modal` context and a visit is proposed to the `default` context. If that visit proposal contains a `replace` action, we don't want to both dismiss the modal screen *and* then `replace` the underlying `default` context screen. That can lead to replacing the `default` context screen that should be preserved, such as the root destination.

The only scenario where we'll both dismiss the modal screen and `replace` is if the underlying `default` context is the same URL as the proposed URL.

### DIFF
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -106,11 +106,12 @@ class NavigationHierarchyController {
             if let visitable = controller as? Visitable {
                 delegate.visit(visitable, on: .main, with: proposal.options)
             }
+            let willContextSwitch = isInModalContext
             navigationController.dismiss(animated: proposal.animated)
             pushOrReplace(on: navigationController,
                           with: controller,
                           via: proposal,
-                          didSwitchToDefaultContext: isInModalContext)
+                          didSwitchToDefaultContext: willContextSwitch)
         case .modal:
             if let visitable = controller as? Visitable {
                 delegate.visit(visitable, on: .modal, with: proposal.options)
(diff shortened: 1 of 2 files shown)


Source/Turbo/Navigator/NavigationHierarchyController.swift
```diff
@@ -94,22 +94,31 @@ class NavigationHierarchyController {
...
+        && !modalNavigationController.isBeingDismissed
+    }
 
     private func navigate(with controller: UIViewController, via proposal: VisitProposal) {
         switch proposal.context {
         case .default:
             if let visitable = controller as? Visitable {
                 delegate.visit(visitable, on: .main, with: proposal.options)
             }
+            let willContextSwitch = isInModalContext
```
jayohms, line 109: For consistency in naming, how about `willDismissModalContext`?

Source/Turbo/Navigator/NavigationHierarchyController.swift
```diff
@@ -119,12 +128,16 @@ class NavigationHierarchyController {
         }
     }
 
-    private func pushOrReplace(on navigationController: UINavigationController, with controller: UIViewController, via proposal: VisitProposal) {
+    private func pushOrReplace(on navigationController: UINavigationController,
+                               with controller: UIViewController,
+                               via proposal: VisitProposal,
+                               didSwitchToDefaultContext: Bool = false) {
```
jayohms, line 134: And here, how about `didDismissModalContext`?

jayohms (approved):
Nice, thanks for the tests!

jayohms:
The corresponding Android PR is ready to go: https://github.com/hotwired/hotwire-native-android/pull/107

Source/Turbo/Navigator/NavigationHierarchyController.swift
olivaresf: What about `willReplaceModalContext` and `didReplaceModalContext`? I like that it's related to the `replace` action, and it reads well in this case.

jayohms: Sounds good to me @olivaresf 👍

svara (approved):
Great! 👏🏻

Issue #82, opened 2025-02-12
VisitableViewController is open but its members are internal access


sfnelson:
`VisitableViewController` and `HotwireNavigationController` are both documented as open classes with the express intent of allowing extension, but `.appearReason` and `.disappearReason` are internal access. This prevents effective subclassing to create custom navigation controllers.

joemasilotti:
Closed via #99.

PR #99, opened 2025-03-10
Fix unremoved screenshot - feedback

svara:
This PR builds on top of https://github.com/hotwired/hotwire-native-ios/pull/91.
It introduces explicit appear and disappear reasons when presenting/dismissing a modal. Even though the new reasons are currently not used anywhere, I think it makes sense to be explicit.

This PR also addresses https://github.com/hotwired/hotwire-native-ios/issues/82 by making the `VisitableViewController`'s `appearReason` and `disappearReason` `open`.

NOTE: Since tabs are a heavily used UI pattern, we should consider implementing a Hotwire Native tab bar controller.

### DIFF
--- a/Source/HotwireNavigationController.swift
+++ b/Source/HotwireNavigationController.swift
@@ -51,6 +51,13 @@ open class HotwireNavigationController: UINavigationController {
         return poppedViewController
     }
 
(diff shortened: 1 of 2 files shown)


joemasilotti (approved):

joemasilotti:
LGTM! I'll close out #91.

PR #101, opened 2025-03-14
Route decision handlers

svara:
This PR brings the concept of a Router and route decision handlers from Android (https://github.com/hotwired/hotwire-native-android/pull/12).

### **Routing**
Before performing an in-app navigation visit, the navigator routes the location URL through the app router. The router loops over each registered route decision handler until it finds a match (the order of the registered route decision handlers determines in what order they're processed). The matched route decision handler must return one of either:

`Router.Decision.navigate`: Informs the navigator to continue performing the in-app navigation visit.
`Router.Decision.cancel`: Informs the navigator to stop the in-app navigation visit. 

The matched route decision handler is given the opportunity to handle the visit by providing its own custom behaviour, such as opening an external browser.

### **Built-in route decision handlers**
There are two built-in route decision handlers that can be used by applications. 

1. **`AppNavigationRouteDecisionHandler`**: Allows location urls on the same domain as `Navigator.Configuration.startLocation` to navigate within the app.
2.  **`BrowserRouteDecisionHandler`**: Allows location urls on a different domain from `Navigator.Configuration.startLocation` to stop in-app navigation and route the URL to an external browser(see `ExternalURLOpeningOption`).

Both `AppNavigationRouteDecisionHandler` and `BrowserRouteDecisionHandler ` are registered by default. The route decision handlers that an application uses can be configured by calling `Hotwire.registerRouteDecisionHandlers([your decision handlers])`. 
Applications can create their own route decision handler implementations by implementing the `RouteDecisionHandler` protocol and registering them as described above.

### Changes

**External URLs**
External urls are now handled by the `BrowserRouteDecisionHandler`. The related old code was removed. Users can still specify how external links should open by setting one of the `ExternalURLOpeningOption` cases to the `HotwireConfig.defaultExternalURLOpeningOption`.

**Navigator Configuration**
The navigator must now be initialized with its configuration -`Navigator.Configuration`. This struct holds a navigator's name and its start location. It is passed to each route decision handler by the Router.

### DIFF
--- a/Demo/NumbersViewController.swift
+++ b/Demo/NumbersViewController.swift
@@ -6,14 +6,14 @@ import UIKit
 final class NumbersViewController: UITableViewController, PathConfigurationIdentifiable {
     static var pathConfigurationIdentifier: String { "numbers" }
 
-    convenience init(url: URL, navigator: Router) {
+    convenience init(url: URL, navigator: NavigationHandler) {
         self.init(nibName: nil, bundle: nil)
         self.url = url
         self.navigator = navigator
     }
 
     private var url: URL!
-    private unowned var navigator: Router?
+    private unowned var navigator: NavigationHandler?
 
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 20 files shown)


svara:
Closed in favour of https://github.com/hotwired/hotwire-native-ios/pull/105.

PR #102, opened 2025-03-19
feat: add request completion to navigation delegate

jupazave:
### Description

This pull request introduces a change that allows us to track whenever a request is finished.

### Background

We initially extended the `HotwireWebViewController` and added a callback on the web side using `WKScriptMessageHandler`. However, we identified that this implementation could be simplified. By leveraging the `SessionDelegate#sessionDidFinishRequest` method, we can handle the event more efficiently.

### Changes

- Added a method `requestDidFinish` to the `NavigatorDelegate` that is called whenever `sessionDidFinishRequest` is fired.

### Testing

- Verified that the `requestDidFinish` method in `NavigatorDelegate` is called appropriately when `sessionDidFinishRequest` is fired.

### Additional Information

If you have any questions or need further clarification, please feel free to reach out.

Thank you for considering this pull request.

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -222,11 +222,10 @@ extension Navigator: SessionDelegate {
 
     public func sessionDidFinishRequest(_ session: Session) {
         guard let url = session.activeVisitable?.visitableURL else { return }
-        
-        delegate.requestDidFinish(at: url)
 
         WKWebsiteDataStore.default().httpCookieStore.getAllCookies { cookies in
             HTTPCookieStorage.shared.setCookies(cookies, for: url, mainDocumentURL: url)
+            self.delegate.requestDidFinish(at: url)
         }
     }
 
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -222,6 +222,8 @@ extension Navigator: SessionDelegate {
 
     public func sessionDidFinishRequest(_ session: Session) {
         guard let url = session.activeVisitable?.visitableURL else { return }
+        
+        delegate.requestDidFinish(at: url)
 
         WKWebsiteDataStore.default().httpCookieStore.getAllCookies { cookies in
             HTTPCookieStorage.shared.setCookies(cookies, for: url, mainDocumentURL: url)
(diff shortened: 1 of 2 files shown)


Source/Turbo/Navigator/NavigatorDelegate.swift
```diff
@@ -33,6 +33,10 @@ public protocol NavigatorDelegate: AnyObject {
     /// Optional. Called after a form finishes a submission.
     /// If not implemented, no action is taken.
     func formSubmissionDidFinish(at url: URL)
+    
+    /// Optional. Called after a request.
```
brol1dev, line 37:
```suggestion
    /// Optional. Called after a request has completed.
```

brol1dev (approved):
Just left one comment but LGTM!

leonvogt:
Nice! I was recently playing around with a launch screen that stays on the screen until the first request is finished.
This change would be very useful for that and would simplify the code quite a bit.

jupazave:
> Nice! I was recently playing around with a launch screen that stays on the screen until the first request is finished. This change would be very useful for that and would simplify the code quite a bit.

Thats one of the this we are trying to do.

Source/Turbo/Navigator/Navigator.swift
```diff
@@ -222,6 +222,8 @@ extension Navigator: SessionDelegate {
 
     public func sessionDidFinishRequest(_ session: Session) {
         guard let url = session.activeVisitable?.visitableURL else { return }
+        
+        delegate.requestDidFinish(at: url)
 
         WKWebsiteDataStore.default().httpCookieStore.getAllCookies { cookies in
             HTTPCookieStorage.shared.setCookies(cookies, for: url, mainDocumentURL: url)
```
joemasilotti, line 229: I think this callback should get fired after the cookies are copied over, when they exist.

jupazave: Our idea is to use it as soon as the render is done, so adding if after the callback will add a little more load time than real one.

jupazave: @joemasilotti are we ok on this?

joemasilotti: According to ChatGPT, I don't think we need to worry much about the time it takes to copy over the cookies.

image

I still think we should wait until the cookies copy before firing the callback. That provides a lot of value for folks who want to do something with a now-authenticated session.

jupazave: @joemasilotti ready to go!

joemasilotti:
Excellent! Thank you. Once we add feature parity to Android we can merge this in for the next release.

svara:
@jayohms and I chatted about this today. We like the proposed functionality, but believe that using async/await fits better. Here's the relevant PR: https://github.com/hotwired/hotwire-native-ios/pull/119.

svara:
Closing in favour of https://github.com/hotwired/hotwire-native-ios/pull/119.

PR #103, opened 2025-03-22
only update the navigation bar title when loading a page, not the tabs

dorianmariecom:
context:

- https://github.com/hotwired/turbo-ios/pull/234
- https://discuss.hotwired.dev/t/hotwire-native-ios-prevent-tab-bar-items-from-having-their-title-changed-on-page-load/6284

### DIFF
--- a/Source/Turbo/Visitable/VisitableViewController.swift
+++ b/Source/Turbo/Visitable/VisitableViewController.swift
@@ -47,7 +47,7 @@ open class VisitableViewController: UIViewController, Visitable {
     // MARK: Visitable
 
     open func visitableDidRender() {
-        title = visitableView.webView?.title
+        navigationItem.title = visitableView.webView?.title
     }
 
     open func showVisitableActivityIndicator() {


joemasilotti:
@jayohms, I'm in favor of this as it is technically what occurs on Android. However, it does break behavior from previous versions.

That said, I usually have to add other, more hacky solutions to my application code to achieve this same result!

svara:
`title = visitableView.webView?.title` updates the view controller’s title property, and `UITabBarController` uses that value for the corresponding tab item if no alternative title is provided.

We have two options:
- Leave the implementation as is. To prevent the tab bar title change one must set the tab bar item's title explicitly.
- Accept the proposed change.

Do we want the web views to drive the tab bar's titles? I'm not in favor of this.

@dorianmariecom I suggest you setup the tab bar controller as follows:
```

    private func loadTabs() {
        tabBarController.viewControllers = Tab.all.map { tab in
            let navigator = Navigator(delegate: self)
            navigator.rootViewController.tabBarItem = UITabBarItem(
                title: tab.title,
                image: UIImage(systemName: tab.imageName),
                selectedImage: nil
            )
            navigators.append(navigator)
            navigator.route(tab.url)
            return navigator.rootViewController
        }
    }

    struct Tab {
        let title: String
        let imageName: String
        let path: String?
        static let baseURL = URL(string: <your url>)!

        var url: URL {
            if let path {
                return Tab.baseURL.appending(path: path)
            }
            return Tab.baseURL
        }

        static let all = {
            var tabs = [
                Tab(title: "Home", imageName: <home image>, path: nil),
                Tab(title: "Programs", imageName: <program image>, path: "programs"),
                Tab(title: "Messages", imageName: <messages image>, path: "messages"),
                Tab(title: "Account", imageName: <account image>, path: "account"),
                Tab(title: "More", imageName: <more image>, path: "more"),
            ]

            return tabs
        }()
    }


```

joemasilotti:
> Do we want the web views to drive the tab bar's titles? I'm not in favor of this.

That's what is currently happening, right? Indirectly, via `title`.

svara:
> > Do we want the web views to drive the tab bar's titles? I'm not in favor of this.
> 
> That's what is currently happening, right? Indirectly, via `title`.

Yes, but only on the first page load.

jayohms:
We discussed this a bit more and this is a nice change. The WebView page title shouldn't have any impact on the tab title — that's unexpected and undesirable in most situations.

jayohms (approved):

dorianmariecom:
I'm officially a contributor :) thanks, and thanks for the fix one less hack in my code

PR #105, opened 2025-03-24
Route decision handlers

svara:
This PR brings the concept of a Router and route decision handlers from Android (https://github.com/hotwired/hotwire-native-android/pull/12).

## Routing
Before performing an in-app navigation visit, the navigator routes the location URL through the app router. The router loops over each registered route decision handler until it finds a match (the order of the registered route decision handlers determines in what order they're processed). The matched route decision handler must return one of either:

`Router.Decision.navigate`: Informs the navigator to continue performing the in-app navigation visit.
`Router.Decision.cancel`: Informs the navigator to stop the in-app navigation visit.

The matched route decision handler is given the opportunity to handle the visit by providing its own custom behaviour, such as opening an external browser.

### Built-in route decision handlers
There are three built-in route decision handlers that can be used by applications.

- `AppNavigationRouteDecisionHandler`: Allows location urls on the same domain as `Navigator.Configuration.startLocation `to navigate within the app.

- `BrowserRouteDecisionHandler`: Allows location urls on a different domain from `Navigator.Configuration.startLocation` to stop in-app navigation and opens the URL via `UIApplication.shared.open`.

- `SafariViewControllerRouteDecisionHandler`: Allows location urls on a different domain from `Navigator.Configuration.startLocation` to stop in-app navigation and opens the URL in `SafariViewController`.

`BrowserRouteDecisionHandler` and `SafariViewControllerRouteDecisionHandler` are mutually exclusive.
`AppNavigationRouteDecisionHandler` and `BrowserRouteDecisionHandler`  are registered by default. The route decision handlers that an application uses can be configured by calling `Hotwire.registerRouteDecisionHandlers([your decision handlers])`.
Applications can create their own route decision handler implementations by implementing the `RouteDecisionHandler `protocol and registering them as described above.

## WebViewPolicyManager
Before a web view performs a navigation, it asks its delegate to make policy decision about whether to allow navigation within the web view. The library delegates the policy decision to the `WebViewPolicyManager`. The policy manager loops over each registered `WebViewPolicyDecisionHandler` instance until it finds a match(the order of the registered policy handlers determines in what order they're processed). The matched policy handler must return one of either:

`WebViewPolicyManager.Decision.cancel`: Informs the web view to cancel the navigation.
`WebViewPolicyManager.Decision.allow`: Informs the web view to allow the navigation.

The matched policy handler is given the opportunity to handle the navigation action (`WKNavigationAction`) by providing its own custom behaviour.

### Built-in route web view policy decision handlers

- `ExternalNavigationWebViewPolicyDecisionHandler`: Intercepts navigation actions where the requested URL should be opened externally.
- `LinkActivatedWebViewPolicyDecisionHandler`: Intercepts navigation actions triggered by link activation in the main frame.
- `NewWindowWebViewPolicyDecisionHandler`: Intercepts navigation actions requesting a new window.
- `ReloadWebViewPolicyDecisionHandler`: Intercepts navigation actions intended to reload the page.

All the above policy handlers are registered by default. The web view policy decision handlers that an application uses can be configured by calling `Hotwire.registerWebViewPolicyDecisionHandlers([your policy decision handlers])`.
Applications can create their web view policy decision handler implementations by implementing the `WebViewPolicyDecisionHandler `protocol and registering them as described above.

## Changes
**External URLs**
External urls are now handled by either `BrowserRouteDecisionHandler` or `SafariViewControllerRouteDecisionHandler`. The related old code was removed. 

**Navigator Configuration**
The navigator must now be initialized with its configuration -`Navigator.Configuration`. This struct holds a navigator's name and its start location. It is passed to each route decision handler by the `Router`.


Source/Turbo/Navigator/Navigator+Configuration.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public extension Navigator {
```
svara, line 3: Following [Android's design](https://github.com/hotwired/hotwire-native-android/blob/main/navigation-fragments/src/main/java/dev/hotwire/navigation/navigator/NavigatorConfiguration.kt).

Source/Turbo/Session/Session.swift
```diff
@@ -452,57 +458,32 @@ extension Session: WebViewDelegate {
...
         log("webViewWebContentProcessDidTerminate")
         delegate?.sessionWebViewProcessDidTerminate(self)
     }
-
-    private func openExternalURL(_ url: URL) {
-        log("openExternalURL", ["url": url])
-        delegate?.session(self, openExternalURL: url)
-    }
-
-    private struct NavigationDecision {
```
svara, line 475: Web view navigation policy decision is handled by the `WebViewPolicyManager` by looping over the provided handlers.

Source/Turbo/Visit/ColdBootVisit.swift
```diff
@@ -65,7 +65,7 @@ extension ColdBootVisit: WKNavigationDelegate {
         if navigationAction.navigationType == .linkActivated {
             decisionHandler(.cancel)
             if let url = navigationAction.request.url {
-                UIApplication.shared.open(url)
+                delegate?.visitDidProposeVisitToLocation(url)
```
svara, line 68: Cold boot link click actions are bubbled up via the delegate and routed through the `Router`.

Source/Turbo/Visit/ColdBootVisit.swift
```diff
@@ -88,7 +88,7 @@ extension ColdBootVisit: WKNavigationDelegate {
         if redirectIsCrossOrigin {
             log("Cross-origin redirect detected: \(location) -> \(url).")
             decisionHandler(.cancel)
-            UIApplication.shared.open(url)
+            delegate?.visitDidProposeVisitToLocation(url)
```
svara, line 91: Cold boot cross origin redirects are bubbled up via the delegate and routed through the `Router`.

Source/Turbo/Navigator/Navigator.swift
```diff
@@ -34,14 +34,25 @@ public class Navigator {
     /// Convenience initializer that doesn't require manually creating `Session` instances.
     /// - Parameters:
     ///   - delegate: _optional:_ delegate to handle custom view controllers
-    public convenience init(delegate: NavigatorDelegate? = nil) {
+    public convenience init(configuration: Navigator.Configuration, delegate: NavigatorDelegate? = nil) {
```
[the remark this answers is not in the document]
jayohms: > The default configuration could handle links like Hotwire Native does currently, before this PR is merged.

@joemasilotti I don't think I follow what this means. All navigation will flow through through the registered `RouteDecisionHandler` instances and the library provides default registered handlers for where things should route. The old hard-coded decisions have been removed.

As far as the `configuration`, we need to pass data to the route decision handlers so they can implement their logic. Is your concern about getting-started aesthetics or something else? I think this is step in the right direction between platforms and having a flexible, consistent `RouteDecisionHandler` API.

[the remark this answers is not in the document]
svara: > Can we provide a default configuration here?

@joemasilotti We can't. Just like on [Android](https://github.com/hotwired/hotwire-native-android/blob/main/demo/src/main/kotlin/dev/hotwire/demo/main/MainActivity.kt), it's the user's responsibility to provide a `configuration`.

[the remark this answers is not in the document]
joemasilotti: Ah, I see now. I missed how this gets initialized.

I'm going to hold off on more feedback until the Demo app is updated. I'm having a hard time wrapping my head around how it all wires up.

>  Is your concern about getting-started aesthetics or something else? I think this is step in the right direction between platforms and having a flexible, consistent `RouteDecisionHandler` API.

I see that now and I agree on the direction.

Source/Turbo/Navigator/Routing/Handlers/SafariViewControllerRouteDecisionHandler.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import SafariServices
+
+/// Opens external URLs via an embedded `SafariViewController` so the user stays in-app.
+/// NOTE: This will silently fail for a URL that's not `http` or `https`.
+final class SafariViewControllerRouteDecisionHandler: RouteDecisionHandler {
+    let name: String = "safari"
+
+    func matches(location: URL,
+                 configuration: Navigator.Configuration) -> Bool {
```
joemasilotti, line 10:
```suggestion
    func matches(location: URL,
                 configuration: Navigator.Configuration) -> Bool {
        guard location.scheme == "http" || location.scheme == "https" else { return false }
```

This would allow `http` and `https` links to open in the in-app browser but let `sms:` and `mailto:` links use `UIApplication.open()` by registering all three handlers. This might even be a better default as it would introduce fewer breaking changes.

```swift
Hotwire.registerRouteDecisionHandlers([
    AppNavigationRouteDecisionHandler(),
    SafariViewControllerRouteDecisionHandler(),
    BrowserRouteDecisionHandler()
])
```

Otherwise, we force developers to create their own decision handler for handling these types of schemes. It also makes `SafariViewControllerRouteDecisionHandler` and `BrowserRouteDecisionHandler` no longer mutually exclusive.

Source/Turbo/Navigator/Navigator.swift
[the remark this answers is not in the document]
joemasilotti: What's the point of `Navigator.Configuration.name`? I see it being used during debug logging, and only twice, but nowhere else.

I get that we are introducing a breaking change with `startLocation`. But can't we make the transition easier by dynamically inferring the class name instead of requiring `name` to be set by the developer?

```diff
diff --git a/Source/Turbo/Navigator/Routing/Router.swift b/Source/Turbo/Navigator/Routing/Router.swift
index 786cd92..b565eb6 100644
--- a/Source/Turbo/Navigator/Routing/Router.swift
+++ b/Source/Turbo/Navigator/Routing/Router.swift
@@ -14,7 +14,7 @@ public final class Router {
                      navigator: Navigator) -> Router.Decision {
         for handler in decisionHandlers {
             if handler.matches(location: location, configuration: configuration) {
-                logger.debug("[Router] handler match found handler: \(handler.name) location: \(location)")
+                logger.debug("[Router] handler match found handler: \(String(describing: type(of: handler))) location: \(location)")
                 return handler.handle(location: location,
                                configuration: configuration,
                                navigator: navigator)
```

```
[Router] handler match found handler: AppNavigationRouteDecisionHandler location: http://localhost:3000/hikes
```

Source/Turbo/Navigator/Routing/Handlers/AppNavigationRouteDecisionHandler.swift
```diff
@@ -0,0 +1,20 @@
...
+
+        return configuration.startLocation.host == location.host
+    }
+
+    func handle(location: URL,
+                configuration: Navigator.Configuration,
+                navigator: Navigator) -> Router.Decision {
+        return .navigate
+    }
+}
```
joemasilotti, line 20: If we want developers to use the handlers then we need to make them public, along with an explicit `init()`. Same goes for the other two handlers.

```suggestion
public final class AppNavigationRouteDecisionHandler: RouteDecisionHandler {
    public let name: String = "app-navigation"

    public init() {}

    public func matches(location: URL,
                 configuration: Navigator.Configuration) -> Bool {
        if #available(iOS 16, *) {
            return configuration.startLocation.host() == location.host()
        }

        return configuration.startLocation.host == location.host
    }

    public func handle(location: URL,
                configuration: Navigator.Configuration,
                navigator: Navigator) -> Router.Decision {
        return .navigate
    }
}
```

Source/Turbo/Navigator/Navigator.swift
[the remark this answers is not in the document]
svara: The main idea for requiring the `name` is to make it easy to write a decision handler that supports switching tabs.

Source/Turbo/Navigator/Routing/Handlers/AppNavigationRouteDecisionHandler.swift
svara: Good point!

Source/Turbo/Navigator/Routing/Handlers/SafariViewControllerRouteDecisionHandler.swift
svara: I like the proposed change. In this case, `BrowserRouteDecisionHandler ` has become more generic.
I suggest we rename it to `SystemNavigationRouteDecisionHandler` or similar.

jayohms: I like this too 👍

joemasilotti: Yes! I like that a lot. I thought Browser vs. Safari was a little ambiguous but couldn't put my finger on why.

jayohms (approved):
Excellent work on this @svara! And love the tests 🙌

PR #108, opened 2025-03-28
Reflect current web view page location in `VisitableViewController`

svara:
This PR adapts the `VisitableViewController` to reflect the current web view page location and matches the reload behaviour with Android.

**Context**
Rails/Hotwire allows a page location change(via redirect or Turbo Frames) without a new visit proposal.
Since the native view controller (`VisitableViewController`) is created with the original URL this change is never reflected (see https://github.com/hotwired/hotwire-native-ios/pull/77).

**Changes**
- `VisitableViewController ` now exposes a `currentVisitableURL` which reflects the associated web view's URL.
- `Visit` uses the `currentVisitableURL` when initialized.

NOTE: The backstack handling hasn't changed. It still relies on the `VisitableViewController`'s  original URL-`visitableURL`.

**Resolving web view's URL**
The `currentVisitableURL` is derived by leveraging an internal `VisitableLocationState` - an enum with three cases:
	•	`resolved`: Indicates that the page has finished loading. In this state, `currentVisitableURL` returns the URL from the web view if available, or falls back to the original `visitableURL`.
	•	`unresolved(URL)`: Represents the initial state before the page has been loaded. Here, the provided original URL is returned.
	•	`deactivated(URL)`: Captures the state when the web view has been deactivated. In this case, the stored URL is used as the current location.

The `resolveVisitableLocation()` function uses these states to determine the correct URL for `currentVisitableURL`. This mechanism ensures that the view controller consistently reflects the current web view's URL throughout its lifecycle, whether it’s still loading, fully loaded, or deactivated.

### DIFF
--- a/Source/Turbo/Visitable/Visitable.swift
+++ b/Source/Turbo/Visitable/Visitable.swift
@@ -14,7 +14,7 @@ public protocol Visitable: AnyObject {
     var visitableViewController: UIViewController { get }
     var visitableDelegate: VisitableDelegate? { get set }
     var visitableView: VisitableView { get }
-    var visitableURL: URL { get }
+    var initialVisitableURL: URL { get }
     var currentVisitableURL: URL { get }
 
     func visitableDidRender()
--- a/Source/Turbo/Visitable/Visitable.swift
+++ b/Source/Turbo/Visitable/Visitable.swift
@@ -14,13 +14,15 @@ public protocol Visitable: AnyObject {
     var visitableViewController: UIViewController { get }
     var visitableDelegate: VisitableDelegate? { get set }
     var visitableView: VisitableView! { get }
-    var visitableURL: URL! { get }
+    var visitableURL: URL { get }
+    var currentVisitableURL: URL { get }
 
     func visitableDidRender()
     func showVisitableActivityIndicator()
     func hideVisitableActivityIndicator()
 
     func visitableDidActivateWebView(_ webView: WKWebView)
+    func visitableWillDeactivateWebView()
     func visitableDidDeactivateWebView()
 }
 
@@ -51,6 +53,7 @@ extension Visitable {
     }
 
     func deactivateVisitableWebView() {
+        visitableWillDeactivateWebView()
         visitableView.deactivateWebView()
         visitableDidDeactivateWebView()
     }
--- a/Source/Turbo/Visitable/Visitable.swift
+++ b/Source/Turbo/Visitable/Visitable.swift
@@ -13,7 +13,7 @@ public protocol VisitableDelegate: AnyObject {
 public protocol Visitable: AnyObject {
     var visitableViewController: UIViewController { get }
     var visitableDelegate: VisitableDelegate? { get set }
-    var visitableView: VisitableView! { get }
+    var visitableView: VisitableView { get }
     var visitableURL: URL { get }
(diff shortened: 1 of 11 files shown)


Source/Turbo/Visitable/VisitableViewController.swift
```diff
@@ -3,15 +3,23 @@ import WebKit
...
-        self.visitableURL = url
+    public init(url: URL) {
+        visitableURL = url
+        visitableLocationState = .unresolved(url)
+        super.init(nibName: nil, bundle: nil)
     }
 
+    required public init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
```
jayohms, line 21: What is the purpose of this?

Source/Turbo/Visitable/VisitableViewController.swift
```diff
@@ -62,13 +71,26 @@ open class VisitableViewController: UIViewController, Visitable {
...
+
     open func visitableDidDeactivateWebView() {
         // No-op
     }
 
+    // MARK: Private
+    enum VisitableLocationState {
+        case resolved
+        case unresolved(URL)
```
jayohms, line 85: How about calling the `unresolved` state `initialized` instead? To me, `unresolved` sounds like an unexpected or undesirable state.

jayohms (commented):
Very nice @svara, thank you 👍

Source/Turbo/Visitable/VisitableViewController.swift
svara: This is a side effect of [removing the explicitly unwrapped optional](https://github.com/hotwired/hotwire-native-ios/pull/108/files#diff-759e47cd198ad8d4977502c4edc279888c7267183391930550628872e6fad79fL16-R18) on the  `visitableURL`. The `visitableURL` must now be set before calling super.

As for the purpose:
This initializer is required because `VisitableViewController` is a subclass of `UIViewController`, which conforms to `NSCoding`. When subclassing `UIViewController`, we are expected to implement the initializer `init?(coder:)` even if we don’t intend to support initialization from a storyboard or nib. By implementing it this way, we are explicitly stating that this view controller should only be instantiated programmatically.

Source/Turbo/Visitable/VisitableViewController.swift
svara: Renamed as suggested.

Source/Turbo/Visitable/Visitable.swift
```diff
@@ -13,14 +13,16 @@ public protocol VisitableDelegate: AnyObject {
 public protocol Visitable: AnyObject {
     var visitableViewController: UIViewController { get }
     var visitableDelegate: VisitableDelegate? { get set }
-    var visitableView: VisitableView! { get }
-    var visitableURL: URL! { get }
+    var visitableView: VisitableView { get }
+    var visitableURL: URL { get }
```
joemasilotti, line 17: Would it be helpful to rename this to `initialVisitableURL`? Having `visitableURL` and `currentVisitableURL` makes it hard for me to understand what the difference is.

svara: Yes, makes sense.

jayohms (approved):
Nice work @svara 👍

PR #109, opened 2025-03-28
Hotwire tab view controller

svara:
This PR builds on top of https://github.com/hotwired/hotwire-native-ios/pull/87 and adds a standalone Hotwire tab view controller.


joemasilotti:
I'd like to see `HotwireTabBarController` implement the `Router` protocol. A naive solution is routing everything to the active navigator.

```swift
class HotwireTabBarController: UITabBarController, Router {
    // ...

    open func route(_ url: URL) {
        activeNavigator.route(url)
    }

    open func route(_ proposal: VisitProposal) {
        activeNavigator.route(proposal)
    }
}
```

Keeping these functions in the class definition, instead of an extension, allows us to mark them as `open`. Then a developer is free to customize the behavior as they see fit.

```swift
class MyTabBarController: HotwireTabBarController {
    override func route(_ url: URL) {
        // Custom behavior.
    }

    override func route(_ proposal: VisitProposal) {
        // Custom behavior.
    }
}
```

Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -0,0 +1,86 @@
...
+        return navigator.rootViewController
+    }
+}
+
+/// Represents a tab in the `HotwireTabBarController`.
+public struct HotwireTab: Hashable {
+    /// The title of the tab.
+    public let title: String
+    /// The image the tab item uses.
+    public let image: UIImage
```
joemasilotti, line 70: This could technically be optional since `UITabBarItem` doesn't require it.

Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -0,0 +1,86 @@
...
+    ///The image the tab item uses when the user selects it.
+    ///If you don’t provide `selectedImage`, the `image` is used for both selection states.
+    public let selectedImage: UIImage?
+    /// The URL associated with the tab, used for routing.
+    public let url: URL
+
+    public init(title: String,
+                image: UIImage,
+                selectedImage: UIImage?,
```
joemasilotti, line 79:
```suggestion
                image: UIImage? = nil,
                selectedImage: UIImage? = nil,
```

PR #113, opened 2025-04-05
Identify navigator when requesting handling by delegate

olivaresf:
This PR addresses an issue where a `NavigatorDelegate` may be handling multiple navigation stacks and needs to distinguish between navigators when a proposal is requested.

⚠️ This is a breaking change since it changes `handle(proposal:)` to `handle(proposal:from:)` ⚠️


Source/Turbo/Navigator/NavigatorDelegate.swift
```diff
@@ -12,8 +12,9 @@ public protocol NavigatorDelegate: AnyObject {
     ///    - term `reject`: No changes to navigation occur.
     ///
     /// - Parameter proposal: `VisitProposal` navigation destination
+    /// - Parameter from: the `Navigator` receiving the proposal
     /// - Returns:`ProposalResult` - how to react to the visit proposal
-    func handle(proposal: VisitProposal) -> ProposalResult
+    func handle(proposal: VisitProposal, from: Navigator) -> ProposalResult
```
joemasilotti, line 17: Spitballing here, but we *could* leave the old function and mark it as deprecated. Then, in `Navigator.controller(for:)`, we could call both functions.

svara: We'll soon launch 1.2 which will contain breaking changes. I suggest we include this breaking change too.

PR #117, opened 2025-04-10
Hotwire tab view controller feedback

joemasilotti:
Hey @svara, this PR implements some feedback I have for #109.

I'll comment inline with my reasoning for each change - let me know if anything doesn't make sense!

### DIFF
--- a/Source/Turbo/ViewControllers/HotwireTabBarController.swift
+++ b/Source/Turbo/ViewControllers/HotwireTabBarController.swift
@@ -5,6 +5,16 @@ import UIKit
 /// This controller loads tabs defined by `HotwireTab` and configures each one with its own `Navigator`.
 /// The currently selected tab's navigator is exposed via the `activeNavigator` property.
 open class HotwireTabBarController: UITabBarController, Router {
+    public init(navigatorDelegate: NavigatorDelegate? = nil) {
+        self.navigatorDelegate = navigatorDelegate
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        fatalError("Use init(navigatorDelegate:) instead.")
+    }
+
     /// The active navigator corresponding to the currently selected tab.
     ///
     /// - Returns: A `Navigator` instance for the currently selected tab.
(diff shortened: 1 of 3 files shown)


Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -4,7 +4,17 @@ import UIKit
 ///
 /// This controller loads tabs defined by `HotwireTab` and configures each one with its own `Navigator`.
 /// The currently selected tab's navigator is exposed via the `activeNavigator` property.
-open class HotwireTabBarController: UITabBarController {
+open class HotwireTabBarController: UITabBarController, Router {
+    public init(navigatorDelegate: NavigatorDelegate? = nil) {
+        self.navigatorDelegate = navigatorDelegate
+        super.init(nibName: nil, bundle: nil)
+    }
```
joemasilotti, line 11: This separates the loading of the tabs with the assigning of the delegate. I'm using this approach to assign a delegate once then dynamically update/load tabs later, from somewhere else in the codebase.

Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -25,17 +35,31 @@ open class HotwireTabBarController: UITabBarController {
...
     /// This method assigns the provided tabs to the controller and sets up each tab's view controller.
-    public func load(_ tabs: [HotwireTab], navigatorDelegate: NavigatorDelegate? = nil) {
+    public func load(_ tabs: [HotwireTab]) {
         hotwireTabs = tabs
         viewControllers = tabs.map {
             setupViewControllerForTab($0, navigatorDelegate: navigatorDelegate)
         }
+        navigatorsByTab.forEach { tab, navigator in
+            navigator.route(tab.url)
+        }
```
joemasilotti, line 45: This ensures that the `viewControllers` property is set before routing any URLs. Before `viewControllers` is set `selectedIndex` is always `NSNotFound` which will cause a `fatalError` when accessing `activeNavigator`.

Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -25,17 +35,31 @@ open class HotwireTabBarController: UITabBarController {
...
+    }
+
+    // MARK: Router
+
+    open func route(_ url: URL) {
+        activeNavigator.route(url)
+    }
+
+    open func route(_ proposal: VisitProposal) {
+        activeNavigator.route(proposal)
```
joemasilotti, line 55: We talked about this [here](https://github.com/hotwired/hotwire-native-ios/pull/109#issuecomment-2774447757).

Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -76,7 +99,7 @@ public struct HotwireTab: Hashable {
 
     public init(title: String,
                 image: UIImage,
-                selectedImage: UIImage?,
+                selectedImage: UIImage? = nil,
```
joemasilotti, line 102: Providing a default value for a non-required parameter cleans up the code a bit.

svara (approved):
Great improvements @joemasilotti! 👏🏻

PR #120, opened 2025-04-19
Add internal parameter name to new function

joemasilotti:
This adds an internal parameter name to the new "handle" function signature. I think this makes it a bit easier to understand when implementing the delegate function.


Source/Turbo/Navigator/NavigatorDelegate.swift
```diff
@@ -14,7 +14,7 @@ public protocol NavigatorDelegate: AnyObject {
     /// - Parameter proposal: `VisitProposal` navigation destination
     /// - Parameter from: the `Navigator` receiving the proposal
     /// - Returns:`ProposalResult` - how to react to the visit proposal
-    func handle(proposal: VisitProposal, from: Navigator) -> ProposalResult
+    func handle(proposal: VisitProposal, from navigator: Navigator) -> ProposalResult
```
joemasilotti, line 17: This is the only "actual" change - everything else propagates from it.

svara (approved):
I like it! 👌

PR #121, opened 2025-04-19
Notify bridge destination when a component is initialized

svara:
This PR adds a function to `BridgeDestination` to get notified when a component is initialized.


joemasilotti (approved):

jayohms (approved):
Looks great, thanks!

PR #125, opened 2025-04-23
Use `weak` instead of `unowned` references for delegates

svara:
This PR refactors the code by using `weak` instead of `unowned` references for delegates.

**Context**
Using `unowned` assumes that the referenced object will always exist, which may not be the case. If the referenced object is accessed after it has been deallocated, the app crashes.

The most common source of crashes is within `BridgeComponent`s due to their unpredictable lifecycles. By using a weak reference instead, we avoid this crash and throw and error.

### DIFF
--- a/Tests/Bridge/BridgeComponentTests.swift
+++ b/Tests/Bridge/BridgeComponentTests.swift
@@ -3,6 +3,7 @@ import HotwireNative
 import WebKit
 import XCTest
 
+@MainActor
 class BridgeComponentTest: XCTestCase {
     private var delegate: BridgeDelegateSpy!
     private var destination: AppBridgeDestination!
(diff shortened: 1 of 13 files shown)


joemasilotti (approved):
TIL!

Tests/Bridge/BridgeComponentTests.swift
```diff
@@ -40,75 +36,35 @@ class BridgeComponentTest: XCTestCase {
...
-    func test_replyToReceivedMessageSucceeds() async throws {
-        let success = try await component.reply(to: "connect")
-
-        XCTAssertTrue(success)
-        XCTAssertTrue(delegate.replyWithMessageWasCalled)
-        XCTAssertEqual(delegate.replyWithMessageArg, message)
-    }
-
-    @MainActor
-    func test_replyToReceivedMessageWithACodableObjectSucceeds() async throws {
```
svara, line 61: We used the same name for async and non-async functions. I noticed that non-async tests were not running, thus I moved them in a separate file.

Issue #122, opened 2025-04-21
Crash in Authentication Challenge Handler


carolinagalvan:
**Description**
Our iOS app is experiencing crashes in the `didReceiveAuthenticationChallenge` method of Hotwire Native's Navigator class. The crash occurs when the Navigator attempts to access its delegate after the delegate has been deallocated, causing a fatal error in the Swift runtime due to an unsafe unowned reference access.

**Stack Trace**
```
0   libsystem_kernel.dylib           __pthread_kill + 8
1   libsystem_pthread.dylib          pthread_kill + 268
2   libsystem_c.dylib                abort + 124
3   libswiftCore.dylib               swift::fatalErrorv(unsigned int, char const*, char*) + 144
4   libswiftCore.dylib               swift::fatalError(unsigned int, char const*) + 32
5   libswiftCore.dylib               swift::swift_abortRetainUnowned(void const*) + 32
6   libswiftCore.dylib               swift_unknownObjectUnownedLoadStrong + 80
7   our-app                          Navigator.session(_:didReceiveAuthenticationChallenge:completionHandler:) + 68
8   our-app                          Session.visit(_:didReceiveAuthenticationChallenge:completionHandler:) + 96
9   our-app                          protocol witness for VisitDelegate.visit(_:didReceiveAuthenticationChallenge:completionHandler:) + 96
10  our-app                          specialized ColdBootVisit.webView(_:didReceive:completionHandler:) + 164
11  our-app                          ColdBootVisit.webView(_:didReceive:completionHandler:) + 20
12  our-app                          @objc ColdBootVisit.webView(_:didReceive:completionHandler:) + 128
13  WebKit                           WebKit::NavigationState::NavigationClient::didReceiveAuthenticationChallenge + 440
```

**Problem Analysis**
The crash occurs in the `Navigator.session(_:didReceiveAuthenticationChallenge:completionHandler:)` method when it tries to call a method on its delegate. The specific error `swift_abortRetainUnowned` indicates that an unowned reference is being accessed after the referenced object has been deallocated.

In the Navigator class, it appears the delegate is held as an unowned reference:
```
public func session(_ session: Session, didReceiveAuthenticationChallenge challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
    delegate.didReceiveAuthenticationChallenge(challenge, completionHandler: completionHandler)
}
```

**Reproduction Steps**
[...]

PR #131, opened 2025-05-01
Fix NavigationDelegateTests

delonnewman:
While investigating this issue https://github.com/hotwired/hotwire-native-ios/issues/122#issuecomment-2845371427 I noticed that `NavigationDelegateTests` wasn't executing. It seems that this was happening because it was a subclass of `Navigator` rather than a subclass of `XCTestCase`. After changing that and setting up the required state the test is being executed and passing.


joemasilotti:
Thanks for calling this out @delonnewman! I updated the test to align it with what is being tested, that the default view controller is being used.

PR #132, opened 2025-05-02
Fix Navigator delegate issues by using a strong reference?

delonnewman:
I'm having a similar issue as @carolinagalvan describes here: https://github.com/hotwired/hotwire-native-ios/issues/122#issue-3009073996. Updating to version 1.2.0 hasn't fixed the problem for me.

If I pass a delegate to the navigator the webview crashes with an `IdleExit` warning. When I change the `delegate` attribute in `Navigator` to a strong reference this issue goes away. I'm not certain of the broader implications of this change.

Device: iPhone 16
iOS Version: iOS 18.2

This code reproduces the issue in the demo app:
```swift
import HotwireNative
import UIKit

let rootURL = URL(string: "https://hotwire-native-demo.dev")!

class SceneDelegate: UIResponder, UIWindowSceneDelegate {

    var window: UIWindow?
    private let navigator = Navigator(
        configuration: Navigator.Configuration(name: "main", startLocation: rootURL),
        delegate: NavigationDelegate()
    )

    func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
        window?.rootViewController = navigator.rootViewController
        navigator.start()
    }
}

class NavigationDelegate: NavigatorDelegate {
    func handle(proposal: VisitProposal, from navigator: Navigator) -> ProposalResult {
        debugPrint(proposal)
        return .accept
    }
}
```

#### IdleExit Warning

Image

#### Crashed WebView

Image

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -8,7 +8,7 @@ class DefaultNavigatorDelegate: NSObject, NavigatorDelegate {}
 /// Handles navigation to new URLs using the following rules:
 /// [Navigator Handled Flows](https://native.hotwired.dev/reference/navigation)
 public class Navigator {
-    public weak var delegate: NavigatorDelegate?
+    public var delegate: NavigatorDelegate?
 
     public var rootViewController: UINavigationController { hierarchyController.navigationController }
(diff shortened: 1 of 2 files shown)


delonnewman:
Got it, thank you!

Issue #53, opened 2024-12-04
Turbo visit with replace action flashes elements - same-page with query params


karlentwistle:
Hey 👋

Thanks for creating and sharing such a brilliant project. My colleagues and I have had an absolute blast using it to make a native app at work.

We have encountered an issue very similar to [hotwired/turbo-ios#136](https://github.com/hotwired/turbo-ios/issues/136), and we're unsure how to circumvent it.

We have these three pages on our website:

- https://www.biggerpockets.com/forums
- https://www.biggerpockets.com/forums?sort=trending
- https://www.biggerpockets.com/forums?sort=unanswered

We've added these in our native application like so:

<img width="500" src="https://github.com/user-attachments/assets/a42089ba-1341-46f5-95cf-29f18326f2c3" />

Since we don't want Turbo to "advance" when the links are clicked, we've configured them via `data-turbo-action="replace"`.

This prevents the page from advancing (to the right) but causes the screen to "flash".

To demonstrate the problem, I've set up a temporary fork of the [turbo-native-demo](https://github.com/hotwired/turbo-native-demo), which should be accessible via:

```diff
diff --git a/Demo/Demo.swift b/Demo/Demo.swift
index 7e261e3..4428bfc 100644
--- a/Demo/Demo.swift
+++ b/Demo/Demo.swift
@@ -1,7 +1,7 @@
 import Foundation

 struct Demo {
-    static let remote = URL(string: "https://hotwire-native-demo.dev/")!
+    static let remote = URL(string: "https://hotwire-native-issue-demo-ef14973e7c91.herokuapp.com/")!
     static let local = URL(string: "http://localhost:45678")!

     /// Update this to choose which demo is run
```

I added a navigation menu with Features and What's New, also using `data-turbo-action="replace"`, and clicking between the navigation links causes the same flashing to occur.

https://github.com/user-attachments/assets/1956867c-bff8-4648-ac0f-65f11445a281

I was wondering if this is expected behaviour or if you have any words of wisdom on how we can avoid this from happening.

Thanks very much.

joemasilotti:
What happens if you add the [morphing](https://turbo.hotwired.dev/handbook/page_refreshes.html) tag to `<head>`?

```
<meta name="turbo-refresh-method" content="morph">
```

karlentwistle:
[...]

PR #143, opened 2025-06-18
URL comparison improvements

svara:


jayohms (approved):
Lovely, thanks @svara!

joemasilotti:
Does this have any effect on #53?

karlentwistle:
> Does this have any effect on #53?

When I saw this PR land, I rushed to test out https://github.com/hotwired/hotwire-native-ios/issues/53 with `data-turbo-action="replace"` removed from the hyperlinks and the path configuration applied instead. Sadly, it appears to have no impact, and the flashing issue remains. 

I think the scope of https://github.com/hotwired/hotwire-native-ios/issues/53 can now be expanded to:
- `data-turbo-action="replace"` on hyperlinks
- `"query_string_presentation": "replace"` on path configuration

PR #150, opened 2025-07-30
Safely cast jsonData to InternalMessageData

joshmcarthur:
This pull request logs a warning when a jsonObject is passed which does not conform to the InternalMessageData type, otherwise retaining the existing behaviour of defaulting to an empty object.

This assists users of the library who might pass a Codable object in as data, and then observe that the value they pass is dropped, without changing any other behaviour.

I personally ran into this attempting to pass an array of result objects in a message reply - because the `as?` conversion fails, the original data was silently dropped.

I'm not generally a Swift developer, so what I've produced might not be idiomatic. I had also hoped to add an assertion covering the logger message, but I couldn't find a way to capture log messages without reimplementing a logger.

### DIFF
--- a/Source/Bridge/InternalMessage.swift
+++ b/Source/Bridge/InternalMessage.swift
@@ -21,7 +21,9 @@ struct InternalMessage {
     }
     
     init(from message: Message) {
-        let data = (message.jsonData.jsonObject() as? InternalMessageData) ?? [:]
+        let jsonObject = message.jsonData.jsonObject()
+        let data = InternalMessage.safeExtractData(from: jsonObject)
+
         self.init(id: message.id,
                   component: message.component,
                   event: message.event,
@@ -46,7 +48,7 @@ struct InternalMessage {
             return nil
         }
         
-        let data = (jsonObject[CodingKeys.data.rawValue] as? InternalMessageData) ?? [:]
+        let data = InternalMessage.safeExtractData(from: jsonObject[CodingKeys.data.rawValue])
         
         self.init(id: id,
                   component: component,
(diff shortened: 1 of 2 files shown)


joemasilotti:
I'm not sure I follow what the original issue is here. Can you share some code of where this will help?

joemasilotti:
I think I understand what you're getting at. Just to make sure I'm not missing anything, you're *not* using a `Codable` `struct` to **reply** to the message, right? For example, like in the [`MenuComponent` from the demo app](https://github.com/hotwired/hotwire-native-ios/blob/main/Demo/Bridge/MenuComponent.swift)?

joshmcarthur:
I was using a `Codable` `struct`, but I was trying to reply with an array of items to the message. 

Something analogous to (from `MenuComponent`):

```swift
private extension MenuComponent {
 

    struct Item: Ccodable {
        let title: String
        let index: Int
    }

  private func onItemSelected(item: Item) {
      reply(
          to: Event.display.rawValue,
          with: [Item(index: 0, title: "Item A")]
      )
  }
} 
```

[`InternalMessageData`](https://github.com/hotwired/hotwire-native-ios/blob/main/Source/Bridge/InternalMessage.swift#L4) is expected to be a string-keyed map, so an array of `Codable` structs (or any other primitive) doesn't cause any build errors, but the [`as?` cast to `InternalMessageData`](https://github.com/hotwired/hotwire-native-ios/blob/main/Source/Bridge/InternalMessage.swift#L24) means that any value like this is silently converted to `[:]`, meaning the message received back in the `BridgeComponent` controller ends up being `{}`. 

Thanks for sticking with me!

joemasilotti:
Got it, thanks! I'm not convinced this needs to be in the library, though. Always using a `Codable`, as the API expects, avoids this.

PR #154, opened 2025-08-13
Pop presenting controller if redirected to modal

### LINKED ISSUE #112: Redirect to Modal Shows Different Screens Between iOS and Android
Dismissing a redirected modal behaves unexpectedly.

**To Reproduce**

* Launch the Demo app (tested on `main` and `1.1.3`)
* Click on "Intercept unauthorized access"
* After modal loads click Done
* A blank screen appears

**Expected**
The Android demo app seems to nicely show the previous page instead. My expectation is that the iOS app would work like the Android app and dismiss to the previously loaded screen. Example videos are below:

**iOS**
Xcode 16.3
iOS Simulator - iPhone 16 Plus - 18.4

https://github.com/user-attachments/assets/2256e0af-7c40-465c-9bb2-b5c2fa35b36d

**Android**
Android Studio Meerkat | 2024.3.1 Patch 1
Pixel 6 Pro API 35

https://github.com/user-attachments/assets/d0889e22-22ca-49ef-87d5-11ad4ee833bf

joemasilotti:
This PR handles the issue described in #112.

Previously, when an endpoint redirected from a non-modal screen to a modal screen, the presenting view controller was left on the main stack. Empty.

This PR checks for the `.replace` navigation action and if it is the first modal screen on the stack, pops the underlying (redirected _from_) view controller on the main stack.

I tested a few use-cases here but I'm sure there are some flows that I missed. Any help finding these would be greatly appreciated!

Fixes #112

### DIFF
--- a/Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
+++ b/Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
@@ -6,7 +6,7 @@ public extension VisitProposal {
     }
 
     var presentation: Navigation.Presentation {
-        options.action == .replace ? .replace : properties.presentation
+        properties.presentation
     }
 
     var modalStyle: Navigation.ModalStyle {
--- a/Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
+++ b/Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
@@ -6,7 +6,7 @@ public extension VisitProposal {
     }
 
     var presentation: Navigation.Presentation {
-        properties.presentation
+        options.action == .replace ? .replace : properties.presentation
     }
 
     var modalStyle: Navigation.ModalStyle {
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -126,6 +126,10 @@ class NavigationHierarchyController {
                 modalNavigationController.setViewControllers([controller], animated: proposal.animated)
                 modalNavigationController.setModalPresentationStyle(via: proposal)
                 navigationController.present(modalNavigationController, animated: proposal.animated)
+
+                if proposal.options.response?.redirected == true {
+                    navigationController.popViewController(animated: false)
+                }
             }
         }
     }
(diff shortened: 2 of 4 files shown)


Source/Turbo/Navigator/NavigationHierarchyController.swift
```diff
@@ -188,6 +188,10 @@ class NavigationHierarchyController {
                 modalNavigationController.setViewControllers([controller], animated: false)
                 modalNavigationController.setModalPresentationStyle(via: proposal)
                 navigationController.present(modalNavigationController, animated: proposal.animated)
+
+                if proposal.presentation == .replace {
+                    navigationController.popViewController(animated: false)
```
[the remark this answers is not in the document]
joemasilotti: `popViewController(animated:)` no-ops if there's only a single view controller on the stack. Ignore.

Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
```diff
@@ -6,7 +6,7 @@ public extension VisitProposal {
     }
 
     var presentation: Navigation.Presentation {
-        properties.presentation
+        options.action == .replace ? .replace : properties.presentation
```
svara, line 9: I'm not leaning towards this. 
We have this nice [callback](https://github.com/hotwired/hotwire-native-ios/blob/595dba37c4918afd0e2d470109bc0c356d6e9afe/Source/Turbo/Navigator/Navigator.swift#L178) for cross origin redirects. 

We can detect if a visit proposal is a redirect by inspecting the options dict. I'm wondering if we should leverage that. @jayohms What do you think?

```
"options": {
    action = replace;
    response =     {
        redirected = 1;
        responseHTML = "<!--...

```

joemasilotti: Ah, great call! I just pushed another commit - what do you think about that approach?

svara: I like it! 
I've got some more feedback so I went ahead and opened [this feedback PR](https://github.com/hotwired/hotwire-native-ios/pull/160). 
I think we should handle the redirect in the same way we handle the cross origin redirect - popping the current destination from the backstack before routing to the new visit proposal.

svara (approved):
Nicely done!

Issue #144, opened 2025-06-24
CSRF Token Invalidated When Deep Linking to Modal Route (/posts/new) Due to Out-of-Order Parallel Requests


rachelgraves:
Versions tested: 1.1.3, 1.2.1

**Summary**
When deep-linking into a modal route (/posts/new) via a push notification, Hotwire Native sends multiple parallel requests. One to the intended modal page (/posts/new) and another to the parent route (e.g., /dashboard). If /dashboard responds after /posts/new, it appears to overwrite the CSRF token set by the modal page, causing form submissions on /posts/new to fail.

**Steps to Reproduce**

- Configure your routing such that `/posts/new` is presented modally.
- Trigger a deep link into `/posts/new` (e.g., via push notification).
- Observe that Hotwire Native makes two XHR requests:
- One to `/posts/new`
- One to its parent page (e.g., `/dashboard`)
- Watch the order in which the responses return. (I used https://proxyman.com/ios)

This bug may not manifest if `/posts/new` is faster than `/dashboard`, making it intermittent and harder to detect.
Reproducing the issue may require forcing the response from /dashboard to be slower than `/posts/new`.
It also seems more likely to happen when the app is cold booting.

**Expected Behaviour**
The page at `/posts/new` should load modally with a valid CSRF token. Submitting its form should work regardless of the timing of other requests.

**Actual Behaviour**
If the parent page (`/dashboard`) responds after `/posts/new`, it appears to overwrite the cookies, including the CSRF token.

Consequently, submitting the form on `/posts/new` fails due to an invalid or missing CSRF token.

**Hypothesis**
Browsers may be applying cookie changes from XHR responses as soon as they arrive, regardless of whether the associated JavaScript has awaited them or not. If this is the case, parallel requests could cause race conditions in cookie-setting behaviour—an issue especially relevant to CSRF token handling.

**Proposed Expectations / Feature Request**

Ideally, the Hotwire Native client could:

Either await or synchronise requests such that CSRF-relevant responses are processed in a known-safe order;

Or support a mode where modal page loads do not trigger automatic parallel requests to parent contexts unless explicitly required.

It would be ideal if browsers only applied cookie changes from XHRs once those promises were awaited—though this may not be realistic.

**Notes**

I'm seeing multiple requests to `/dashboard` and I'm not sure why

joemasilotti:
The invalid CSRF issue can be reproduced by loading a modal URL immediately after loading the initial tabs in the demo app:

[...]

PR #157, opened 2025-08-19
Lazy load tabs

joemasilotti:
This PR changes the behavior of how tabs are loaded via `HotwireTabBarController`. Before, all tabs are loaded as soon as `load(_:)` is called. Now, only the first tab is loaded. The other tabs load only after navigating to that tab.

This change removes a [race condition for CSRF protection](#144) if root screens have `<form>` tags. It also reduces server load on app launch (only one request is made instead of up to five).

### DIFF
--- a/Source/Turbo/ViewControllers/HotwireTabBarController.swift
+++ b/Source/Turbo/ViewControllers/HotwireTabBarController.swift
@@ -8,6 +8,7 @@ open class HotwireTabBarController: UITabBarController, NavigationHandler {
     public init(navigatorDelegate: NavigatorDelegate? = nil) {
         self.navigatorDelegate = navigatorDelegate
         super.init(nibName: nil, bundle: nil)
+        delegate = self
     }
 
     @available(*, unavailable)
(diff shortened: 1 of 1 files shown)


Source/Turbo/ViewControllers/HotwireTabBarController.swift
```diff
@@ -92,6 +93,15 @@ open class HotwireTabBarController: UITabBarController, NavigationHandler {
     }
 }
 
+extension HotwireTabBarController: UITabBarControllerDelegate {
+    public func tabBarController(_ tabBarController: UITabBarController, didSelect viewController: UIViewController) {
+        guard activeNavigator.rootViewController.viewControllers.isEmpty else { return }
+
+        let tab = hotwireTabs[selectedIndex]
+        activeNavigator.route(tab.url)
```
svara, line 101:
```suggestion
        activeNavigator.start()
```

joemasilotti: Oh, much cleaner! Thanks @svara.

nbelzer:
@joemasilotti Really great add-on! What would you recommend in case we do want to pre-load all or certain tabs for a better user experience?

I noticed that the `navigatorsByTab` are private so I'm unable to simply loop over them and call `start()` when launching the application.

joemasilotti:
What about #194?

PR #159, opened 2025-08-19
Configuration to hide the tab bar on push

### LINKED ISSUE #141: Config / setup to hide the tab bar on view push
Sorry if this belongs to the forums instead, I am looking for a way to hide the tab bar when a user clicks on a link in the app. Some apps out there hide the tab bar on view push. 

I've played around with subclassing `HotwireWebViewController` and setting `self.hidesBottomBarWhenPushed = true` on `init` but this prevents the tab bar from being shown in the first place.

joemasilotti:
```swift
Hotwire.config.hidesTabBarWhenPushed = true
```

When disabled (default behavior):

https://github.com/user-attachments/assets/6046a5be-750d-493b-980a-6048317cc974

When enabled:

https://github.com/user-attachments/assets/7ebde285-e05d-470f-81b0-9eab03fe61e4

Should fix #141.

### DIFF
--- a/Source/HotwireConfig.swift
+++ b/Source/HotwireConfig.swift
@@ -22,6 +22,9 @@ public struct HotwireConfig {
     /// Sets the back button display mode of `HotwireWebViewController`.
     public var backButtonDisplayMode = UINavigationItem.BackButtonDisplayMode.default
 
+    /// Set to true to only show the tab bar on the root screens.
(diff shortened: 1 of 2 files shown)


Jacoblab1:
Nice! We actually had implemented our own version of this. Will be great to remove that code and switch to this config option

svara (approved):
Nicely done 👍🏻

PR #160, opened 2025-08-20
Modal redirects feedback

svara:
This is a feedback PR for https://github.com/hotwired/hotwire-native-ios/pull/154.

Changes

- Exposes `VisitResponse` initializer.
- Adds convenience `isRedirect` var to `VisitProposal` extension.
- Handles redirect in the same way as cross origin redirect is handled - popping the current destination from the backstack.

### DIFF
--- a/Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
+++ b/Source/Turbo/Navigator/Extensions/VisitProposalExtension.swift
@@ -29,6 +29,10 @@ public extension VisitProposal {
         properties.animated
     }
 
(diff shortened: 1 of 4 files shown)


svara:
> Perhaps naive, but disabling the pop animation when in the main stack brings back the old behavior:

@joemasilotti  Good catch! Have you seen this happen anywhere else?

svara:
@joemasilotti Thanks. I've further tweaked the code to animate the pop only when the active session is modal and the visit is proposed on the default context. This covers the case when a redirect is proposed inside a modal flow.

PR #165, opened 2025-08-21
Fix same page `restore` visits when returning to a web screen from a native screen

svara:
This is the iOS PR corresponding to the [Android one](https://github.com/hotwired/hotwire-native-android/pull/160).


jayohms (approved):
Thanks @svara!

jayohms:
After further testing, this approach does not work in the "default" context. The intention was to match the Android behavior, but since iOS doesn't have an issue (like Android) reconnecting bridge components when restoring the same page after a native visit, I'm closing this.

PR #171, opened 2025-09-08
Descriptive errors

zoejessica:
Three changes to increase the descriptiveness and accuracy of some HotwireNative errors:

- conforms `RedirectHandlerError` to `LocalizedError` so we can pass on an exact description of the underlying error to the client
-  ensures that we use the precise status code passed from the server when a redirect failed, instead of always assuming a `0` status code. 
- fail with a `contentTypeMismatch` error (instead of status code `0`) if the the url response is neither http nor a blob

### DIFF
--- a/Source/Turbo/Networking/RedirectHandler.swift
+++ b/Source/Turbo/Networking/RedirectHandler.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-enum RedirectHandlerError: Error {
+public enum RedirectHandlerError: LocalizedError {
     case requestFailed(Error)
     case responseValidationFailed(reason: ResponseValidationFailureReason)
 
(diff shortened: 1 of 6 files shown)


svara (commented):
This is a good start. However, I'd like us to consolidate all the errors (maybe in a follow up PR?). Similar to what [Android](https://github.com/hotwired/hotwire-native-android/tree/main/core/src/main/kotlin/dev/hotwire/core/turbo/errors) has. 

I imagine an enum like:
```
enum HotwireNativeError: Error {
    case redirectionFailed(reason: RedirectFailureReason)
    ...
}
```

`HotwireNativeError` would encompass a few different types of errors, each with their own associated reasons.

zoejessica:
Closed in favour of (and incorporated in) https://github.com/hotwired/hotwire-native-ios/pull/215

PR #175, opened 2025-09-20
Ensure consistent URL path handling across iOS versions

augustorsouza:
Added a new method to retrieve URL paths using `URLComponents`, preventing the backslash removal that occurs with URL.path on iOS 15.

Validated in iOS 16+, but for iOS 15.5 that is the one I was able to run i my mac only validated with unit tests since the Demo application didn't run.


augustorsouza:
@joemasilotti could you review or assign a reviewer to this PR, please? :)

joemasilotti:
I posted feedback in #211.

PR #182, opened 2025-10-01
Delete cookies before syncing them from WebView to HTTPCookieStorage

ibrahima:
This PR deletes cookies before syncing them from the WebView session to HTTPCookieStorage. Otherwise, when cookies are deleted on the WebView side of things, they may never get deleted from HTTPCookieStorage, which can lead to unexpected behavior.

I ran into an issue with our app where we were using URLSession.shared to make a request to log in to the app, and then copying cookies from HTTPCookieStorage to the WebView. On log out, we delete cookies relevant to the previously logged in user, and this does delete them from the WebView. But when we log back in to a different user account, the cookies from the previous user get restored. I looked around and it _seems_ like deleted cookies in the WebView never get deleted from HTTPCookieStorage, though I don't know for sure if my understanding is correct. But I was able to solve it in my app by deleting all cookies on logout. But I think it makes most sense to fix it here, upstream, so that this issue doesn't bite other people. It seems like the logically correct thing to do to ensure that cookies are always in sync with what the WebView session has.

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -229,6 +229,11 @@ extension Navigator: SessionDelegate {
         guard let url = session.activeVisitable?.initialVisitableURL else { return }
 
         Task { @MainActor in
+            // To ensure deletions in the WebView are mirrored, first remove any shared cookies
+            // for this URL, then set the cookies read from WKWebView.
+            let storage = HTTPCookieStorage.shared
+            (storage.cookies(for: url) ?? []).forEach { storage.deleteCookie($0) }
+
             let cookies = await WKWebsiteDataStore.default().httpCookieStore.allCookies()
             HTTPCookieStorage.shared.setCookies(cookies, for: url, mainDocumentURL: url)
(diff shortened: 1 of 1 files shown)


joemasilotti:
I think this should be handled in your application code, not the library.

If a developer manually sets some cookies in `HTTPCookieStorage` (outside of the ones from the web view) then this PR will delete them on every page load.

ibrahima:
Hmm maybe, but if a developer manually sets cookies that overlap with ones from the webview they would already get overridden. It seems a little surprising that it syncs cookie creation from the backend but not deletion. At least for me it felt like a footgun that setting cookies in Rails gets propagated to the http cookie storage but deleting cookies in Rails does not.

Perhaps the safest thing would be to only delete cookies that were deleted on the server side based on the headers, but that would be a bit more complicated (not sure if we could do it here).

joemasilotti:
> It seems a little surprising that it syncs cookie creation from the backend but not deletion. 

Then your server should send the cookie with an expired expiration date and the existing code will delete it.

PR #194, opened 2025-10-16
Add configuration to lazy load tabs

joemasilotti:
This also means that 1.3.0 introduces one fewer breaking change!


tonydehnke:
Any chance we cold get this merged into a release version?

Source/HotwireConfig.swift
```diff
@@ -29,6 +29,9 @@ public struct HotwireConfig {
     /// Set to `true` to fade content when performing a `replace` visit.
     public var animateReplaceActions = false
 
+    /// Set to `true` to load each tab after it is selected (instead of on app launch).
+    public var lazyLoadTabs = false
+
```
svara: To keep the current behavior we should set this to `true`. 
cc @jayohms @joemasilotti

svara: I set it to `true` by default to keep the current behavior.

svara (approved):

PR #195, opened 2025-10-21
Allow universal links to open in app within SafariViewController - feedback

svara:
This PR improves https://github.com/hotwired/hotwire-native-ios/pull/191 by opening universal links via the active `UIWindowScene` instead of `UIApplication.shared.open`.


joemasilotti (approved):
Ah, nice! Good call.

PR #211, opened 2025-12-01
Url path preserving backslash feedback

### LINKED ISSUE #114: Path configuration: Trailing slash in path handled inconsistently between iOS and Android
Noticed a minor inconsistency between iOS and Android in how path configurations are handled/parsed. This is by no means a critical bug, but the difference in behavior might confuse developers who start with iOS and then try to use the same configuration for their Android app.

On iOS, the following snippet will cover all paths with and without a trailing `/` - e.g. both `https://www.example.com/task/add` and `https://www.example.com/task/add/` or both `https://www.example.com/task/123/update` and `https://www.example.com/task/123/update/` will open in a modal:

```
"rules": [
    {
      "patterns": [
        "/add$",
        "/update$",
      ],
      "properties": {
        "context": "modal",
        "pull_to_refresh_enabled": false
      }
    }
]
```

On Android, the config file needs to explicitly cover both cases:
```
"rules": [
    {
      "patterns": [
        "/add$",
        "/add/$",
        "/update$",
        "/update/$",
      ],
      "properties": {
        "context": "modal",
        "pull_to_refresh_enabled": false
      }
    }
]
```
(Alternatively, making the trailing slash optional - `"/add/?$"` - also should work on both Android and iOS according to my limited testing.)

The way this is handled on Android probably (?) is more accurate from a technical point of view (`$` to only match end of string).

joemasilotti:
Closes #114.


joemasilotti:
@augustorsouza, here's my take on this. What do you think?

augustorsouza:
I loved it :)

LGTM

augustorsouza (approved):

svara (approved):

PR #215, opened 2025-12-31
Structured errors HotwireNativeError enum

alpkeser:
Following up on @svara's suggestion [here](https://github.com/hotwired/hotwire-native-ios/pull/171#pullrequestreview-3256609796) and aligning the error types with https://github.com/hotwired/turbo-android/pull/317, this PR replaces `TurboError` with unified `HotwireNativeError` type.

HotwireNativeError introduces three distinct cases:
- `WebError`: Represents network issues or `URLError`
- `HttpError`: Represents server response with an error code (4xx, 5xx)
- `LoadError`: Represents custom `Turbo.js` errors when it's failed to load

Similar to Android counterpart this will be a breaking change as we're updating delegate method signatures.


Source/Turbo/Errors/LoadError.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// The server returned an invalid content type (non-HTML response).
+    case contentTypeMismatch
+
+    /// The server returned a malformed or unexpected response.
+    case invalidResponse
+
+    public var description: String {
+        switch self {
+        case .notPresent: return "Turbo Not Present"
+        case .notReady: return "Turbo Not Ready"
```
alpkeser, line 20: This is not in use since the check is missing on the js/bridge side. I suppose we can match it with Android @jayohms?

Source/Turbo/Session/Session.swift
```diff
@@ -424,16 +424,43 @@ extension Session: WebViewDelegate {
...
+    private func errorFromStatusCode(_ statusCode: Int) -> HotwireNativeError {
+        switch statusCode {
+        case -2:
+            return .load(.contentTypeMismatch)
+        case ...0:
+            return .web(WebError.from(turboStatusCode: statusCode))
+        default:
+            return .http(HttpError.from(statusCode: statusCode))
+        }
+    }
```
zoejessica: This logic is repeated in `JavascriptVisit` - maybe this should become an `init` on `HotwireNativeError` instead? `init(turboJSStatusCode: Int) { ... } `?

Source/Turbo/Errors/WebError.swift
```diff
@@ -0,0 +1,121 @@
...
+        return [.notConnectedToInternet, .networkConnectionLost].contains(code)
+    }
+
+    /// Whether the request timed out.
+    public var isTimeout: Bool {
+        if let code = urlError?.code {
+            return code == .timedOut
+        }
+        // Turbo.js status code -1 = timeout
+        return errorCode == -1
```
zoejessica: Can we check `errorCode` directly here, since it's already setup at `init` with the `urlError.code` if it exists?

Source/Turbo/Visit/JavaScriptVisit.swift
```diff
@@ -60,9 +60,23 @@ extension JavaScriptVisit: WebViewVisitDelegate {
...
+        // - 0 = network failure, -1 = timeout, -2 = content type mismatch
+        let error: HotwireNativeError
+        switch statusCode {
+        case -2:
+            error = .load(.contentTypeMismatch)
+        case ...0:
+            error = .web(WebError.from(turboStatusCode: statusCode))
+        default:
+            error = .http(HttpError.from(statusCode: statusCode))
+        }
```
zoejessica: See duplication comment above.

Source/Turbo/Errors/WebError.swift
```diff
@@ -0,0 +1,121 @@
...
+        switch turboStatusCode {
+        case 0:
+            description = "Network failure"
+        case -1:
+            description = "Timeout"
+        default:
+            description = "Network error"
+        }
+        return WebError(errorCode: turboStatusCode, description: description)
+    }
```
zoejessica: What do you think about making this non-public so it's only accessed by a specialized init on HotwireNativeError for the Turbo status codes? Reading the rest of the code it looks incorrect to initialize this with anything <0 (that's an http error) or a -2 (that's a content type mismatch error).

zoejessica (approved):

Source/Turbo/Errors/LoadError.swift
jayohms: It looks like we should probably add a similar callback to the native side like Android. It'd go here: https://github.com/hotwired/hotwire-native-ios/blob/main/Source/Turbo/WebView/turbo.js#L233-L235

Android implementation:
https://github.com/hotwired/hotwire-native-android/blob/588d69c124f2baf7fe605c91fd96d67896e6b534/core/src/main/assets/js/turbo.js#L266

Source/Turbo/Errors/HttpError.swift
```diff
@@ -0,0 +1,178 @@
+import Foundation
+
+/// Errors representing HTTP status codes received from the server.
+public enum HttpError: LocalizedError, Equatable {
```
olivaresf: Can we rename this to HTTPError?

Source/Turbo/Errors/WebError.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Whether the server could not be reached.
+    public var isConnectionError: Bool {
+        guard let code = urlError?.code else { return false }
+        return [.cannotFindHost, .cannotConnectToHost, .dnsLookupFailed].contains(code)
+    }
+
+    /// Whether this is an SSL/TLS error.
+    public var isSslError: Bool {
```
olivaresf: Can we rename this to isSSLError?

Source/Turbo/Errors/WebError.swift
```diff
@@ -0,0 +1,117 @@
...
+    public init(errorCode: Int, description: String?) {
+        self.urlError = nil
+        self.errorCode = errorCode
+        self.description = description ?? "Network Error"
+    }
+
+    // MARK: - Factory Methods
+
+    /// Creates a WebError from a URLError.
+    public static func from(_ urlError: URLError) -> WebError {
```
olivaresf: Is there a reason we need this when we have `init(urlError:)`?

Source/Turbo/Networking/RedirectHandler.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-enum RedirectHandlerError: Error {
+public enum RedirectHandlerError: LocalizedError {
```
olivaresf: Why are we making this public?

Source/Turbo/Errors/LoadError.swift
zoejessica: Got it. If I'm following the Android implementation correctly, this time out path fires both errors in quick succession: first .notReady via `turboIsReady(isReady = false)`, and then .notPresent via `.turboFailedToLoad()`. I'll implement the same on iOS.

Source/Turbo/Networking/RedirectHandler.swift
zoejessica: Good catch, leftover from https://github.com/hotwired/hotwire-native-ios/pull/215/commits/059073c7d39b78f57c049c215cedf90577911ebe before the HotwireNativeError was introduced - we can mark internal 👍

Source/Turbo/Errors/WebError.swift
zoejessica: No I don't think this is adding anything 👍

svara:
Closing in favor of https://github.com/hotwired/hotwire-native-ios/pull/231.

PR #225, opened 2026-01-30
Handle requests for new windows by proposing a new visit

zoejessica:
Changes default Hotwire Native behaviour to automatically start a new visit to the requested url when the `WKUIDelegate` receives a request to create a new web view for a navigation.

The motivation here was to handle links requesting new windows that are embedded in an iframe seamlessly through HotwireNative navigation, rather than discarding them. For example, routing the links in a Youtube embed out to the native Youtube app. Previously, iOS clients would have to subclass `WKUIController` and implement their own handling of this `WKUIDelegate` method:
```swift
func webView(_ webView: WKWebView, createWebViewWith configuration: WKWebViewConfiguration, for navigationAction: WKNavigationAction, windowFeatures: WKWindowFeatures) -> WKWebView?
```

The new behaviour proposed here matches the implementation in hotwire-native-android in [`HotwireWebChromeClient`](https://github.com/hotwired/hotwire-native-android/blob/588d69c124f2baf7fe605c91fd96d67896e6b534/core/src/main/kotlin/dev/hotwire/core/turbo/webview/HotwireWebChromeClient.kt#L74).

We attempt to filter out common blank urls before calling the delegate method. The Android implementation avoids this step by using `requestFocusNodeHref`, which isn't available to us. 

If `Navigator` is not used, the default implementation of the new delegate method does nothing, so this is a non-breaking change.


Source/Turbo/Navigator/WKUIController.swift
```diff
@@ -30,4 +35,32 @@ open class WKUIController: NSObject, WKUIDelegate {
...
+
+private extension URL {
+    var isRoutableNewWindowURL: Bool {
+        guard !absoluteString.isEmpty else {
+            return false
+        }
+
+        switch absoluteString {
+        case "about:blank", "about:srcdoc":
+            return false
```
jayohms: Could it be better to check for `http(s)://` prefix instead or would that be too restrictive for other protocols like `mailto`, `tel`, etc? I don't know how likely those are for `iframe` content.

jayohms (commented):
Thank you @zoejessica! Left one question...

Source/Turbo/Navigator/WKUIController.swift
zoejessica: Yes that was also a path I considered, and didn't do, precisely of `mailto` uses, and because Android doesn't check the scheme either I think? `requestFocusNodeHref` would just return the raw link url without worrying about the scheme, if I have understood its use properly 🤔.

jayohms: Yeah, correct we'll let any scheme through. All good, I was mostly wondering b/c we don't have to worry about `about:blank` type strings coming though. The one other option here maybe would be to use a Url model/parser to validate that its a valid url (with any scheme) and then you don't have to worry about explicitly covering specific conditions?

zoejessica: @svara [suggested](https://github.com/basecamp/bc3-ios/pull/1249#discussion_r2753355409) an improvement I like: we could actually go through the whole stack of `WebViewPolicyDecisionHandler`s , by directly routing through the existing web view. So instead of going to `navigator.route(url)` and introducing a new delegate method, we could directly invoke `webView.load(navigationAction.request)`. 

(As he points out the only change here is that the `targetFrame` of the navigation action becomes the existing main webView rather than nil, but that's the same info as we lose going straight to the `Navigator` anyway. )

That would mean we _could_ filter out invalid schemes in `ExternalNavigationWebViewPolicyDecisionHandler` - if we detect one of those, we'd handle but cancel the navigation. (And clients could modify the invalid list by providing their own handler in `Hotwire.registerWebViewPolicyDecisionHandlers`).

jayohms: Sounds pretty good to me, thanks @zoejessica!

svara: I ported the changes from https://github.com/basecamp/bc3-ios/pull/1249.

svara (approved):

PR #228, opened 2026-02-12
Adds an implementation of the LoadError.notReady error

zoejessica:
Matches Android’s implementation, as suggested [here](https://github.com/hotwired/hotwire-native-ios/pull/215#discussion_r2800064150). When a page timeout fires, the framework will now post a `LoadError.notReady` error first, before going through the `pageLoadFailed` flow.


svara:
Closing in favor of https://github.com/hotwired/hotwire-native-ios/pull/231.

PR #231, opened 2026-02-13
Replace TurboError with structured error types and improve cold boot retry

olivaresf:
This PR replaces the generic `TurboError` enum with a structured `HotwireNativeError` type hierarchy that provides better error categorization, richer context, and smarter recovery behavior.

 **New Error Types**
- Add HotwireNativeError as the unified error type with three cases: .http(HTTPError), .web(WebError), and .load(LoadError).
- Add HTTPError with exhaustive ClientError (4xx) and ServerError (5xx) cases, each with an isRetryable property (only 408, 429, 503, 504).
- Add WebError wrapping URLError with semantic helpers: isOffline, isTimeout, isConnectionError, isSSLError.
- Add LoadError for Turbo.js initialization failures: .notPresent, .notReady, .contentTypeMismatch, .invalidResponse.

**RedirectHandler changes** 
 - Remove HTTP status code validation from RedirectHandler — any HTTP response proves the server is reachable.
 - Add 30-second timeout to RedirectHandler requests.
 - Cold boot retry on both reachable servers (.noRedirect) and network failures (.requestFailed), limited to one retry per visit.
 - Make retry handler conditional on `error.isRetryable` in Navigator.

**Update JS Bridge**
 - Update the JS bridge to split visitRequestFailedWithStatusCode into HTTP and non-HTTP paths, passing status codes to the native side.

===

### Cold-boot retry (handled by hotwire-native)

When a Turbo.js visit fails with a non-HTTP status (error 0, timeout, etc.), the library now:

1. Sends the same request via `RedirectHandler` to check server reachability
2. If the server responds (any status code) → **cold-boot retry** (once)
3. If the server is unreachable (timeout, DNS, connection refused) → **cold-boot retry** (once), then show the error


Source/Turbo/WebView/WebViewBridge.swift
```diff
@@ -112,8 +112,13 @@ extension WebViewBridge: ScriptMessageHandlerDelegate {
         switch message.name {
         case .pageLoaded:
             pageLoadDelegate?.webView(self, didLoadPageWithRestorationIdentifier: message.restorationIdentifier!)
+        case .turboIsReady:
+            let isReady = message.data["isReady"] as? Bool ?? false
```
alpkeser: Nitpick: Like others, we could define a property for this on `ScriptMessage`.

Source/Turbo/Networking/RedirectHandler.swift
```diff
@@ -21,7 +34,8 @@ struct RedirectHandler {
 
     func resolve(location: URL) async throws -> Result {
         do {
-            let request = URLRequest(url: location)
+            var request = URLRequest(url: location)
+            request.timeoutInterval = 30
```
zoejessica: As discussed with @olivaresf , if this lands in a HN release we'll need to a) investigate if we can make the turbo timeout and url request timeouts concurrent rather than seemingly additive and b) probably make this a config value for clients.

Source/Turbo/WebView/WebViewBridge.swift
svara: Fixed in [75b2e32](https://github.com/hotwired/hotwire-native-ios/pull/231/commits/75b2e32b9d5ef69e202b320ef39f3f5accf1bd32).

Source/Turbo/Networking/RedirectHandler.swift
svara: Addressed in [c441e60](https://github.com/hotwired/hotwire-native-ios/pull/231/commits/c441e60349f5b6987ce56fb4159cb6131ec9593a).

svara (approved):
👌

PR #237, opened 2026-05-04
Use PathIdentifiable for navigation identity checks

alpkeser:
`visitingSamePage` and `visitingPreviousPage` fall back to view-controller type equality when the candidate isn't a `Visitable` / `VisitableViewController`. Two instances of the same custom class with different URLs are then indistinguishable, causing an incorrect pop or replace instead of a push.

Adds a `PathIdentifiable` protocol exposing `initialVisitableURL`. `Visitable` inherits from it (existing conformers are unchanged), and both checks now prefer URL identity over type identity. Custom native view controllers can opt in by conforming to `PathIdentifiable`.

### DIFF
--- a/Source/Turbo/Navigator/Helpers/PathIdentifiable.swift
+++ b/Source/Turbo/Navigator/Helpers/PathIdentifiable.swift
@@ -1,0 +1,15 @@
+import UIKit
+
+/// A view controller may conform to `PathIdentifiable` to expose the URL it represents.
+///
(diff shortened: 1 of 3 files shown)


svara (approved):
This is ok as a temporary fix(not merging to main).
Long term, we should aim to have something similar to Android where the library takes care of assigning the destination URL automatically.

svara:
Closing in favor of #248.

Issue #234, opened 2026-03-06
Snapshot cache not cleared after form submission in modal


leonvogt:
There seems to be an issue where Hotwire Native iOS doesn't invalidate the snapshot cache after submitting a form inside a modal.

---

**Current behavior of Turbo JS:**

After non-safe (POST/PUT/DELETE/PATCH) form submissions, Turbo clears the cache of all snapshots and refetches the page when navigating back to it. This ensures that users see the most up-to-date content after performing actions that modify data.    
Example:

- Navigate from /page_1 to /page_2
- Submit form and get redirected to /page_1

The cached snapshot of /page_1 should be invalidated, and the new content should be visible when visiting /page_1 again.

---

**Current behavior of Hotwire Native iOS:**

- Navigate from /page_1 to /page_2 (where /page_2 is presented as a modal)
- Submit form and get redirected to /page_1

Expected: The cache of /page_1 should be invalidated, and the new content should be visible when visiting /page_1 again.
Actual: The cache of /page_1 is not invalidated, and the old content is still visible when visiting /page_1 again.

Reproduction steps can be found in the [invalidated-cache-bug](https://github.com/hotwired/hotwire-native-demo/compare/main...leonvogt:hotwire-native-demo:invalidated-cache-bug) of my [hotwire-native-demo fork](https://github.com/leonvogt/hotwire-native-demo/tree/invalidated-cache-bug).

https://github.com/user-attachments/assets/c6fabca3-01c8-41a8-93fb-56b7a0a61f48

---

**Notes:** 

- On Android, it works as expected 
- When following a link back to /page_1 instead of submitting a form, /page_1 reloads every time. In my head, it would be the other way around: GET requests to the previous page might show the cached version, but redirects after a non-safe form submission should always refetch the page.
- When /page_2 is not presented as a modal, it works as expected. After submitting the form, the cache is invalidated and the new content is visible when visiting /page_1 again via the redirect.

https://github.com/user-attachments/assets/ea7365a9-ac24-47ba-b140-f3fc9fee078b

en30:
I ran into what seems to be the same problem, and I noticed that it does not reproduce for me with the changes in this branch:

https://github.com/en30/hotwire-native-ios/tree/en30/modal-visit-fixes

I also tried using that branch with the reproduction repo linked in this issue, and I was not able to reproduce the problem there either.

My branch also includes an unrelated modal redirect fix, but the part that seems relevant here is the stale modal content change.

[...]

PR #240, opened 2026-06-01
Fix refresh_historical_location restoring from snapshot cache

hulloitskai:
## Summary

- `refreshVisitable` in `Navigator.swift` was issuing `action: .restore`, which is Turbo's back-navigation action and prefers a cached snapshot when one exists. That contradicts the "refresh" intent expressed by `refresh_or_redirect_to` / `/refresh_historical_location`: after a non-safe form submission, the underlying page should reflect the new server state, not a snapshot taken before the mutation.
- Switched both branches of `refreshVisitable` to `action: .replace`, which re-fetches and renders the historical location in place without adding to history.

## Repro (before this fix)

1. Navigate from `/page_1` to `/page_2` (in any presentation that ends up using the snapshot cache).
2. Submit a non-safe form on `/page_2`; server returns `refresh_or_redirect_to /page_1` → `/refresh_historical_location`.
3. First submission appears correct only because the default session's snapshot cache happened to be empty.
4. Navigate to `/page_2` again, submit again. Now `/page_1` is rendered from the cached snapshot taken before the first mutation — new content is missing.

Bridge logs show `[Session] visit ... action=restore ... hasCachedSnapshot=1` on the second cycle, confirming the restore path is reusing the cached snapshot.

Related: hotwired/hotwire-native-ios#234.

## Test plan

- [ ] Manual: reproduce the scenario above against an app using `refresh_or_redirect_to` (or any server-side redirect to `/refresh_historical_location`) and confirm `/page_1` shows fresh content on every refresh, not just the first.
- [ ] Existing tests in `Tests/Turbo/Navigator/` still pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -253,9 +253,9 @@ extension Navigator: NavigationHierarchyControllerDelegate {
     func refreshVisitable(navigationStack: NavigationHierarchyController.NavigationStackType, newTopmostVisitable: any Visitable) {
         switch navigationStack {
         case .main:
-            session.visit(newTopmostVisitable, action: .restore)
+            session.visit(newTopmostVisitable, action: .replace)
         case .modal:
-            modalSession.visit(newTopmostVisitable, action: .restore)
+            modalSession.visit(newTopmostVisitable, action: .replace)
         }
     }
 }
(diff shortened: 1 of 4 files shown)


hulloitskai:
Forgot to fix tests, opening new PR.

PR #246, opened 2026-06-30
Introduce Navigating protocol to make decision handlers testable

svara:
## Why

Route and web view policy decision handlers receive a concrete `Navigator`, so unit-testing a custom handler means subclassing `Navigator` and dragging in real `Session` and `WKWebView` instances. Handlers actually touch only a thin slice of `Navigator` (`route`, `reload`, `activeNavigationController`).

This introduces a `Navigating` protocol so handlers depend on an abstraction and can be tested against a lightweight double.

## What

- **New `Navigating` protocol** that refines the existing `NavigationHandler`, exposing the navigation surface handlers actually use: the `route` overloads, `pop`/`clearAll`/`reload`, the root/modal/active navigation controllers, and a new `currentVisitableURL` convenience (default implementation reads the active navigation controller's top visitable).
- `Navigator` conforms with no additional code.
- `RouteDecisionHandler` / `WebViewPolicyDecisionHandler` (and the `Router` / `WebViewPolicyManager` that drive them) now take `Navigating` instead of the concrete `Navigator`.
- `NavigationSpy` is now a lightweight standalone `Navigating` conformer (no `Session`/`WKWebView`), and the decision handler tests use it instead of a real `Navigator`.

## Design notes

- **Surface kept lean.** `session`/`modalSession`, `webkitUIDelegate`, `delegate`, and `start()` are intentionally **not** on the protocol — they're implementation details and would force test doubles to fabricate heavyweight collaborators.
- **All `route` overloads live on `NavigationHandler`.** The full-control `route(_:options:parameters:)` is the requirement; `route(_:)` is a default implementation in terms of it, so conformers (incl. `HotwireTabBarController`) implement only the primitives.

## Breaking changes (source)

Both are minor and apply only to code that **implements** these protocols, not code that calls them:

1. Custom `RouteDecisionHandler` / `WebViewPolicyDecisionHandler` implementations must change their `handle(...)` `navigator` parameter from `Navigator` to `Navigating` (one-line signature change; bodies unchanged). This is the intended enabler for testing them.
2. Adding `route(_:options:parameters:)` as a `NavigationHandler` requirement affects any external type conforming to `NavigationHandler` directly (a consume-not-implement protocol; no known external conformers).

## Testing

The routing handler suites also got noticeably faster now that they no longer spin up real sessions/web views.


zoejessica (approved):
Nicely done @svara, I'm really happy to see this one get merged in! 👏🚀

PR #250, opened 2026-07-06
Move `lazyLoadTabs` from `Hotwire.config` to `HotwireTabBarController` initializer

svara:
### Summary

This PR moves the `lazyLoadTabs` setting from `Hotwire.config` to a `HotwireTabBarController` initializer parameter.

Lazy tab loading was introduced in [#194](https://github.com/hotwired/hotwire-native-ios/pull/194) as a global config option. Since the behavior is specific to `HotwireTabBarController`, it fits better as an initializer parameter than as library-wide configuration, and this aligns the API shape with Android, where [hotwired/hotwire-native-android#202](https://github.com/hotwired/hotwire-native-android/issues/202) exposes `lazyLoadTabs` as a constructor parameter on `HotwireBottomNavigationController`.

### Notes
The defaults intentionally differ between platforms: iOS defaults to true (lazy loading shipped here as the default), while Android defaults to false (eager loading was its released behavior, so lazy loading is opt-in there).

### DIFF
--- a/Demo/AppDelegate.swift
+++ b/Demo/AppDelegate.swift
@@ -42,7 +42,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         Hotwire.config.backButtonDisplayMode = .minimal
         Hotwire.config.showDoneButtonOnModals = true
         Hotwire.config.animateReplaceActions = true
-        Hotwire.config.lazyLoadTabs = true
 #if DEBUG
         Hotwire.config.debugLoggingEnabled = true
 #endif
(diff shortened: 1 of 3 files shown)


joemasilotti:
> The defaults intentionally differ between platforms: iOS defaults to true (lazy loading shipped here as the default), while Android defaults to false (eager loading was its released behavior, so lazy loading is opt-in there).

Should we align that before we release 1.3?

svara:
> > The defaults intentionally differ between platforms: iOS defaults to true (lazy loading shipped here as the default), while Android defaults to false (eager loading was its released behavior, so lazy loading is opt-in there).
> 
> Should we align that before we release 1.3?

Yes. I mistakenly thought we defaulted to true in 1.2.2. Fixed in https://github.com/hotwired/hotwire-native-ios/pull/252.

PR #192, opened 2025-10-15
Don't pop if not redirecting from MODAL session

joemasilotti:
Fixes #190.

Issue #190, opened 2025-10-14
Redirects misbehaving, missing back button or back button goes back to wrong page


dennispaagman:
I was playing with the 1.3 beta and noticed that tabs gets stuck after a redirect.

A few examples with the demo app:

Opening a redirect first, no back button, so the tab is stuck:

https://github.com/user-attachments/assets/3deef58f-b9d0-4215-b3d5-3d41382c2655

I also noticed something strange with redirects after the first page:

Here I open a regular page first, then a redirect (for this I added the redirect navigation item to page, navigations#show in the demo app). There is back button now, but this resets the tab to the first page, not the second page. So it's probably the back button from the previous page. 

https://github.com/user-attachments/assets/cb243f4e-b437-466e-98dc-95691de51846

This all works as expected in 1.2!

PR #251, opened 2026-07-06
Only pop the pre-redirect controller when a redirect crosses contexts

svara:
## Summary

Redirects were popping one view controller too many. This narrows the pop so it only fires when a redirect crosses the `default` ↔ `modal` navigation boundary — fixing the missing / incorrect back button in #190 while preserving the modal-redirect cleanup from #112.

## The bug (#190)

Since the 1.3.0 beta, following an internal redirect could leave the navigation stack wrong:

- **Redirect on/near the first screen** → the tab's root screen is lost and there's no back button.
- **Redirect after navigating** → a back button appears but returns to the wrong screen (one too far back).

Works correctly in 1.2.x.

## Root cause

`session(_:didProposeVisit:)` popped the top controller on **every** redirect (introduced in #160, generalizing the modal fix from #154 / #112):

```swift
if proposal.isRedirect {
    let animatePop = session === modalSession && proposal.context == .default
    pop(animated: animatePop)
}
route(proposal)
```

But Turbo already re-proposes a followed redirect with an `action: "replace"` — it also does `history.replaceState` on the web side, so the redirect never becomes a history entry. Verified in the bundled `turbo.min.js` (`Visit#followRedirect`):

```js
followRedirect(){ … this.response?.redirected &&
  this.adapter.visitProposedToLocation(this.redirectedToLocation,
    { action: "replace", response: this.response, … }) }
```

So `route(proposal)` on its own already **replaces the pre-redirect controller in place** (`pushOrReplace` → `replaceLastViewController`). The extra unconditional `pop` therefore removes a *second* controller — the screen underneath — which is exactly the lost / wrong back button in #190.

The pop is only actually required when the redirect changes navigation **context** (`default` ↔ `modal`): there the pre-redirect controller was pushed onto one stack while `route` targets the other, leaving it orphaned (the original #112 problem).

## The fix

Pop only when the redirect crosses the context boundary:

```swift
if proposal.isRedirect {
    let sessionIsModal = session === modalSession
    let proposalIsModal = proposal.context == .modal
    if sessionIsModal != proposalIsModal {
        // Animate only when receding from the modal session to the default context.
        let animatePop = sessionIsModal && proposal.context == .default
        pop(animated: animatePop)
    }
}
route(proposal)
```

- **Same context** (main → main, modal → modal): no pop; the `replace` action rewrites the pre-redirect controller in place. When the redirect targets the previous screen, `pushOrReplace`'s `visitingPreviousPage` branch performs the **animated** back-pop — restoring the transition that went missing in the beta.
- **Cross context** (default → modal, modal → default): pop the orphaned pre-redirect controller before routing, preserving the #112 fix.

## Behavior

| redirect | before (current beta) | after |
|---|---|---|
| main → main (after a screen) | loses the underlying screen ❌ | replaces in place ✅ |
| main → main (at root) | ok | ok ✅ |
| main → modal (#112) | ok | ok ✅ |
| modal → modal | loses the underlying screen ❌ | replaces in place ✅ |
| modal → main | dismiss + route | dismiss + route ✅ |

## Relationship to #192

#192 also narrows the pop, but keys it off the **session** (`session === modalSession`). A `default → modal` redirect is re-proposed by the **main** session (which performed the request), so #192 skips the pop there and re-introduces #112 — the pre-redirect controller is left orphaned beneath the presented modal. It also still double-pops `modal → modal`. Keying off the **context transition** instead handles all four quadrants.

## Tests

Adds redirect coverage across the full `default` / `modal` matrix to `NavigationHierarchyControllerTests`, driven through the real `SessionDelegate` entry point (`session(_:didProposeVisit:)`) rather than `route(_:)`.

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -175,10 +175,20 @@ public class Navigator {
 extension Navigator: SessionDelegate {
     public func session(_ session: Session, didProposeVisit proposal: VisitProposal) {
         if proposal.isRedirect {
-            // Animate the pop only if we're in the active modal session
-            // and the visit is proposed on the default context.
-            let animatePop = session === modalSession && proposal.context == .default
-            pop(animated: animatePop)
+            // Turbo re-proposes a followed redirect with a `replace` action (it also
+            // performed `history.replaceState` on the web side), so routing the
+            // proposal already rewrites the pre-redirect controller in place. We only
+            // need to pop that controller when the redirect moves across the
+            // default <-> modal boundary, otherwise the
+            // controller is orphaned on the stack it was pushed onto.
+            let sessionIsModal = session === modalSession
+            let proposalIsModal = proposal.context == .modal
+            if sessionIsModal != proposalIsModal {
+                // Animate the pop only when receding from the active modal session
+                // to the default context, matching the back-style transition.
+                let animatePop = sessionIsModal && proposal.context == .default
+                pop(animated: animatePop)
+            }
         }
         route(proposal)
     }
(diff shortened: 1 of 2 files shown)


joemasilotti (approved):
Approved assuming this fixes #192!

PR #260, opened 2026-08-03
Route modal-to-default redirects after the modal dismissal completes

olivaresf:
Since 1453caf, `session(_:didProposeVisit:)` pops the modal for a redirect that crosses from the modal session to the default context and then routes the proposal synchronously. The navigator delegate is therefore consulted while the dismissal transition is still in flight, and delegates that key off presentation state can drop the proposal — in HEY iOS this swallowed the post-form redirect entirely, leaving a stale web view on screen (Screener "Clear all").

This defers routing until the dismissal actually completes: `NavigationHierarchyController.dismissModal(animated:completion:)` dismisses the entire modal stack (dismiss, not pop — multi-controller modal stacks are covered) and invokes its completion immediately when nothing is presented; the modal-to-default redirect proposal is routed from that completion. Delegates now always see a settled hierarchy.

**To verify**
- New tests cover the deferred route for one- and multi-controller modal stacks, asserting the original proposal reaches the delegate only after dismissal completes.
- `TestableNavigationController` now honors the dismiss-completion contract itself (UIKit never finishes a presentation transition in a test bundle, so forwarding the completion to `super` silently dropped it) — that's the `fixup!` commit, to be autosquashed before this leaves draft.
- Full suite: 368/368 on iOS Simulator.

**Compatibility notes**
- Modal-to-default redirects now route after the dismissal animation instead of synchronously — a new async seam for code that relied on the old ordering.
- Default-to-modal and same-context redirects deliberately keep their current order.

Draft pending review and autosquash.

### DIFF
--- a/Source/Turbo/Navigator/NavigationHierarchyController.swift
+++ b/Source/Turbo/Navigator/NavigationHierarchyController.swift
@@ -76,6 +76,15 @@ class NavigationHierarchyController {
         }
     }
 
+    func dismissModal(animated: Bool, completion: (() -> Void)? = nil) {
+        guard navigationController.presentedViewController != nil else {
+            completion?()
+            return
+        }
+
+        navigationController.dismiss(animated: animated, completion: completion)
+    }
+
     func clearAll(animated: Bool) {
         navigationController.dismiss(animated: animated)
         navigationController.popToRootViewController(animated: animated)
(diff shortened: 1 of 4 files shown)


olivaresf:
Mistakenly merged the branch 🤦🏻‍♂️

PR #261, opened 2026-08-03
Route modal-to-default redirects after the modal dismissal completes

olivaresf:
Related: https://app.basecamp.com/2914079/buckets/43795599/card_tables/cards/10152469020
Verified via basecamp/hey-ios#1657, which pins HEY Email to this branch for the manual gate.

Since 1453caf, a modal-to-default redirect routes while the navigator is still dismissing the modal. The delegate gets consulted mid-transition, so a delegate that looks at what's presented sees a half-dismissed hierarchy and can turn the proposal down — that's how HEY iOS lost a post-form redirect, leaving a stale list on screen.

Routing now waits for the dismissal to finish. Other redirect transitions keep their current order.

Replaces #260, reverted because it merged before a consuming app had verified it.

## Reproduction

**On `bridge-delegate-destination-lifecycle`, the bug is reproducible:**
1. In a consuming app whose delegate inspects presentation state (HEY iOS), open a web modal whose form POST redirects to a default-context location — Screener → "Clear all…" → the confirm page.
2. Submit the form. The modal dismisses, but the redirect never routes — the screen beneath still shows the pre-submit list.

**On this branch, it's fixed:**
1. Pin the app to `ab9b9cc` (basecamp/hey-ios#1657 does this) and repeat. The modal dismisses, then the redirect routes — you land on the Imbox, and the Screener is empty on re-entry.

**Without the GUI:** run the `NavigationHierarchyControllerTests` class — `test_redirect_modalToMain_dismissesModalAndRoutesOntoMainStack` holds the dismissal completion and asserts the original proposal routes only after it fires.

### DIFF
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -183,6 +183,14 @@ extension Navigator: SessionDelegate {
             // controller is orphaned on the stack it was pushed onto.
             let sessionIsModal = session === modalSession
             let proposalIsModal = proposal.context == .modal
+
+            if sessionIsModal && !proposalIsModal {
+                hierarchyController.dismissModal(animated: true) { [weak self] in
+                    self?.route(proposal)
+                }
+                return
+            }
+
             if sessionIsModal != proposalIsModal {
                 // Animate the pop only when receding from the active modal session
                 // to the default context, matching the back-style transition.
--- a/Source/Turbo/Navigator/Navigator.swift
+++ b/Source/Turbo/Navigator/Navigator.swift
@@ -184,18 +184,16 @@ extension Navigator: SessionDelegate {
             let sessionIsModal = session === modalSession
             let proposalIsModal = proposal.context == .modal
 
-            if sessionIsModal && !proposalIsModal {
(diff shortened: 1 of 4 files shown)


Source/Turbo/Navigator/Navigator.swift
svara, line 199: One cleanup I would make: after the early modal to default return, the remaining cross context case can only be default to modal, so `animatePop` is always false. A switch would make this more readable.

```
switch (sessionIsModal, proposalIsModal) {
case (true, false):
    hierarchyController.dismissModal(animated: true) { [weak self] in
        self?.route(proposal)
    }
    return
case (false, true):
    pop(animated: false)
default:
    break
}
```

svara (approved):
Great catch 👍🏻
I've left one code style comment.

Source/Turbo/Navigator/Navigator.swift
svara: Addressed in [0392628](https://github.com/hotwired/hotwire-native-ios/pull/261/commits/0392628fbd007c3e7a0c36581e43e70fb88971b8).

PR #263, opened 2026-08-27
Refuse script messages that aren't from the main frame

olivaresf:


olivaresf:
Implementation-only diff: dbc4fc0fec0d2401f0be46015793bfcb29ec07a4..54b3e6e6a7244b1c7c8cf63fcce85eb0299e8b01 — `Source/ScriptMessageHandler.swift`, 1 file changed, 6 insertions(+), 0 deletions. Tests land in a separate commit on top of this range; this range is the implementation read.

Context receipt — acceptance clause built to: "every legitimate bridge component still works; a foreign-origin subframe gets no bridge at all, with a log line saying why." Platform contract of record: `WKScriptMessage.frameInfo` → `WKFrameInfo.isMainFrame` / `.securityOrigin` (all non-optional, verified against the iOS 26.4 SDK WebKit headers rather than cited from memory).

Baseline: existing suite green on this range — 368 tests passed via `xcodebuild test -scheme HotwireNative -destination "name=iPhone 17 Pro"`, the repo's own CI invocation.

PR #265, opened 2026-09-15
Hand the pull-to-refresh control to the scroll view, so the spinner stays centred

devjah:
## Summary

On iOS 27 the pull-to-refresh spinner spins half off the screen's left edge (seen in DOTVOZ Agency and every other fork, since all pin this package).

`VisitableView` added its `UIRefreshControl` to the web view's scroll view as a plain subview and centred it with Auto Layout constraints (centre-X on the visitable view, top on the safe area, a fixed height, no width). A scroll view on iOS 27 adopts a `UIRefreshControl` subview as its own `refreshControl` and sets the frame itself the moment refreshing begins; that frame lands at x = 0 with the zero width the constraints had left it, so the spinner (centred in the control) ends up centred on the screen edge. iOS 26 does not adopt the subview, so the constraints kept winning there.

The control is now handed over through `scrollView.refreshControl`, the documented path since iOS 10, and UIKit keeps it centred above the content on every release. `removeRefreshControl` also clears `scrollView.refreshControl` explicitly: with the old code on iOS 27, the shared web view's scroll view kept the previous screen's control after `deactivateWebView`.

`VisitableViewRefreshControlTests` covers ownership, centring while refreshing, and hand-back on deactivation and on `allowsPullToRefresh = false`. All four fail against the previous implementation on the iOS 27 simulator and pass with this change.

Forks pin this by revision; `purcel-diet-ios` moves first, the template and the other forks follow.


devjah:
Opened against the wrong repository by mistake (meant for my fork). Sorry for the noise.

